PR #3, opened 2024-04-16
ci: add firebase configs

conorsch:
Generated via `firebase init`. Configures project association, so that running `firebase deploy` in the repo root will push the previously built static docroot from `apps/prax-marketing-site/dist/` to the public website: https://praxwallet.com.


apps/prax-marketing-site/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "minifront",
+  "name": "prax-marketing-website",
```
conorsch, line 2: Updated the package name to minimize confusion; I think the new one is correct, override me if not.

turbo.json
```diff
@@ -3,6 +3,7 @@
   "globalDependencies": ["**/.env.*local"],
   "pipeline": {
     "build": {
+      "cache": false,
```
conorsch, line 6: Disabled caching for prod builds, because in my local testing, `rm -rf apps/prax-marketing-site/dist/ && pnpm build` did not regenerate the docroot as I'd expected. Clearly I don't understand something about the turbo/next caching, but rather than debug it, I forced rebuilds every time.

PR #34, opened 2024-06-17
use npm packages

turbocrime:
includes some rebased code that was dropped


TalDerei:
Looks good after green CI + external deps are swapped in

.github/workflows/turbo-ci.yml
```diff
@@ -7,6 +7,9 @@ on:
     branches:
       - main
 
+env:
+  DO_NOT_TRACK: 1
```
turbocrime, line 11: replaces turbo telemetry disable command

.github/workflows/extension-upload.yml
```diff
@@ -13,32 +13,29 @@ on:
...
     steps:
       - uses: actions/checkout@v4
       - id: built
         uses: buildjet/cache@v3
         with:
           path: .turbo
-          key: ${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-built
-          restore-keys: ${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-compiled
+          key: ${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-publish
+          restore-keys: ${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-test
```
turbocrime, line 32: possibly restore cache from a successful test run

TalDerei:
why doesn't locally linking package deps like this work?

Screenshot_2024-06-19_at_12 21 33_PM
Screenshot_2024-06-19_at_12 21 56_PM

turbocrime:
> why doesn't locally linking package deps like this work?

part of this is syncpack, which doesn't know how to compare paths except by a literal string match, and can't compare semver matchers with paths

you may also be having problems with module resolution, which was not addressed by the packaging work, and package builds, which turbo won't assist with or understand for external packages

'link' is really only useful for packages that have no build step, or are a directory essentially equivalent to an unpacked tarball

TalDerei (approved):
confirmed public packages are working!

grod220:
Nice job here! 🎊

PR #42, opened 2024-06-19
Scope approvals to documentId

grod220:


apps/extension/src/listeners/tabs-updated-prax-init.ts
```diff
@@ -3,14 +3,26 @@ import { PraxConnection } from '../message/prax';
 
 // trigger injected-connection-port to init when a known page is loaded.
 chrome.tabs.onUpdated.addListener(
-  (tabId, { status, discarded }, { url }) =>
+  (tabId, { status, discarded, url: changeInfoUrl }, { url: tabUrl }) => {
+    // Unfortunately, no context for tab
+    // void chrome.runtime.getContexts({}).then(console.log);
```
grod220, line 8: @turbocrime, sadly the runtime is only reporting BACKGROUND context here. This means all of the code below is not executing as we expect.
Screenshot 2024-06-20 at 1 21 43 PM

PR #43, opened 2024-06-19
fix(extension): #40: pause ApproveDeny component when out of focus

### LINKED ISSUE #40: Popup countdown focus logic
Currently, the tx & approval countdown in the popup runs regardless if it is in focus or not:

Screenshot 2024-06-18 at 5 31 54 PM

We should update the logic:
- Timer should only countdown if the popup is in focus
- If the user clicks away, the timer should reset

VanishMax:
Closes #40 

When the window is out of focus, approval countdown freezes. When the focus is back, countdown restarts from the beginning


grod220 (approved):
Nice job 👍 ⏲️

PR #44, opened 2024-06-19
ui package parity with web

TalDerei:
currently the ui packages are out of sync, which makes dev confusing. this attempts to maintain some level of _temporary_ parity until the ui package is properly fragmented into `ui/web` and `ui/prax`.


turbocrime (changes requested):
there is no reason to update components that are not used in prax. they should be removed

PR #45, opened 2024-06-20
prax/ui

TalDerei:


packages/tailwind-config/index.ts
```diff
@@ -8,8 +8,8 @@ export default {
     './components/**/*.{ts,tsx}',
     './app/**/*.{ts,tsx}',
     './src/**/*.{ts,tsx}',
-    '../../packages/ui/components/**/*.{ts,tsx}',
     './shared/**/*.{ts,tsx}',
+    './node_modules/@penumbra-zone/ui/**/*.js'
```
VanishMax, line 12: This line makes UI package work correctly with Tailwind.

Why: Tailwind performs JIT compilation to compose and export to a CSS file all used classes. In case of UI package, we have many dynamic classes, which are not parsed without this line change. For example, in `<Button variant='gradient'>` the variant dynamically adds classes and, to process them by JIT, users of UI package need to point Tailwind config to the node_modules.

Of course, it requires users to use Tailwind. If we want to change, we need to update the build process to export the JS files along with a CSS. This file must collect all styles from all components because otherwise users will have to import `button.css`, `table.css`, etc.

TalDerei:
@grod220 this pr follows the suggested compartmentalization defined in the issue description https://github.com/penumbra-zone/web/issues/1336. The corresponding `web/ui` pr is https://github.com/penumbra-zone/web/pull/1342. Should this PR specifically be closed in favor of keeping the UI package in this repo, following L's suggestions?

grod220:
@TalDerei https://github.com/penumbra-zone/web/issues/1336 really just focuses on organizing `@penumbra-zone/ui` so that base level primitives can be exported to anticipated consumers. Radiant has a goal of making updated UI primitives with the dex/block explorer in mind especially (given they don't really look related to penumbra at the moment). 

But the question with this PR is: should Prax also consume things like <Button>, <Card>, etc also from this shared library? And it really comes down to A) what in the new UI components will Prax actually use? will Prax look completely different and therefore the shared library isn't really relevant? and B) how can they be used with security in mind? In short, that probably means:

- For now, we live with the UI package fork (perhaps there's some stuff we can delete in there). Prax UI will be out of sync with minifront + ecosystem base-level components.
- Post new brand guidelines, Radiant will also work on Prax design updates. At that point, it will become clearer what in the shared library that Prax 2.0 should use.
- And then we can either:
   -  Consume those specific base-level components (like buttons) via NPM on a pinned version
   - Or copy them over again

TalDerei:
> @TalDerei [penumbra-zone/web#1336](https://github.com/penumbra-zone/web/issues/1336) really just focuses on organizing `@penumbra-zone/ui` so that base level primitives can be exported to anticipated consumers.

sounds great! closing in favor of follow-up work that will maintain and clean up the UI package fork

PR #50, opened 2024-06-22
hmr script 

TalDerei:
references https://github.com/prax-wallet/web/issues/59

`watch-and-repack.sh` script simulates hot module replacement (**HMR**) by using `fswatch` to detect changes in the `penumbra-zone/web` repository. I think the script can be configured to optionally use the watch capabilities of `tsc`. When changes are detected, the script performs the following actions in sequence:

**1.** triggers `pack-public.sh` and repacks the packages.
**2.** installs the tarballs as dependencies in prax.
**3.** reloads webpack.

this process was tricky to set up because `fswatch` was initially continuously triggering the repack process because the repack itself was causing changes that fswatch picked up, leading to an infinite loop. 

--------------------------------------------------------------------------------------------------------------------------
### to use the prax dev script

(0) enable script permissions in `penumbra-zone/web` with `chmod +x pack-public.sh` and in `prax-wallet/web` with`chmod +x watch-and-repack.sh`.

(1) set the environment variables `PENUMBRA_ZONE_WEB_PATH` and `PRAX_REPO_PATH` with relative paths.

(2) remove npm dependencies from `prax-wallet/apps/extension/package.json` and any other local packages, e.g.
```
    "@penumbra-zone/bech32m": "^6.1.0",
    "@penumbra-zone/client": "^8.0.0",
    "@penumbra-zone/crypto-web": "^5.0.0",
    "@penumbra-zone/getters": "^8.0.0",
    "@penumbra-zone/keys": "^4.1.0",
    "@penumbra-zone/perspective": "^6.0.0",
    "@penumbra-zone/protobuf": "^5.1.0",
    "@penumbra-zone/query": "^6.0.0",
    "@penumbra-zone/services": "^6.0.0",
    "@penumbra-zone/storage": "^6.0.0",
    "@penumbra-zone/transport-chrome": "^4.0.0",
    "@penumbra-zone/transport-dom": "^7.1.0",
    "@penumbra-zone/types": "^9.0.0",
    "@penumbra-zone/wasm": "^9.0.0",
```
(3) execute `pnpm install`.

(4) execute `pnpm run watch-and-repack`, and it will wait for changes in `penumbra-zone/web` before starting the execution process.

--------------------------------------------------------------------------------------------------------------------------
### demo

https://github.com/prax-wallet/web/assets/70081547/8b42aa3c-234a-4d51-99d6-ab200947cc4a


watch-and-repack.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+# Relative paths to the web repositories
+PENUMBRA_ZONE_WEB_PATH="../../web"
```
TalDerei, line 4: devs should modify this relative path

turbocrime: instead of unconditionally clobbering this var, check it if exists and is a valid dir using `readlink`

watch-and-repack.sh
```diff
@@ -0,0 +1,40 @@
...
+PRAX_REPO_PATH="."
+
+# Ensure 'pack-public.sh' is executable
+chmod +x "$PENUMBRA_ZONE_WEB_PATH/pack-public.sh"
```
turbocrime, line 8: don't modify permissions like this

watch-and-repack.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# Remove any existing tarballs from packages in penumbra-zone/web 
+cleanup_tarballs() {
+  find "$PENUMBRA_ZONE_WEB_PATH/packages" -name "*.tgz" -type f -delete
```
turbocrime, line 12: this is unnecessary and dangerous.

the `pack` command will easily clobber any existing tgz.

if the var is empty you will be deleting from root `/packages`. this find may also unexpectedly remove other tgz below its target, such as in `node_modules` (which symlinks out of this tree).

generally avoid deleting anything in shell scripts

watch-and-repack.sh
```diff
@@ -0,0 +1,40 @@
...
+}
+
+# Repack the packages in penumbra-zone/web
+repack() {
+  (cd "$PENUMBRA_ZONE_WEB_PATH" && ./pack-public.sh)
```
turbocrime, line 17: should probably allow other repos to control their own pack behavior, and place output in an external target directory.

watch-and-repack.sh
```diff
@@ -0,0 +1,40 @@
...
+}
+
+# Install dependencies in prax-wallet/web
+install_prax() {
+  (cd "$PRAX_REPO_PATH" && pnpm add -w ../../web/packages/*/penumbra-zone-*-*.tgz)
+}
+
+# Reload webpack
+reload_webpack() {
+  (cd "$PRAX_REPO_PATH" && pkill -f "webpack-dev-server")
```
turbocrime, line 27: pkill is not directory constrained. in scripts you should only kill by known pid or job control.

it should be possible to signal webpack to reload without killing it

turbocrime (changes requested):
some negative things here. we should pair on this

watch-and-repack.sh
TalDerei: script can remove this, and dev can manually set the permission. I just did this for my personal convenience

watch-and-repack.sh
TalDerei: I originally was trying to reload webpack, but changes only seemed to propagate when webpack was killed and then reloaded with pnpm dev

TalDerei:
I think this should just be regarded as a temporary and minimal stop gap solution to help accelerate our dev workflows in the meantime.

watch-and-repack.sh
TalDerei: pkill uses pattern matching under the hood, but not being directory constrained isn't relevant here. It's true that it will terminate any process matching the pattern regardless of the current directory, so if there are multiple webpack dev server instances running, `pkill -f "pnpm run dev"` will terminate all of them. 

I made the pattern unique,` webpack-prax-dev-server` to only terminate the dev server in prax.

watch-and-repack.sh
TalDerei: modified to check env vars

turbocrime: to be more clear - i think it is better to fail entirely if the test fails. the correct location will be incredibly user-specific, and also this default is also not checked the same way as an incoming var, so is likely to fail anyway.

watch-and-repack.sh
TalDerei: Screenshot 2024-06-24 at 3 00 43 PM

will work on the webpack because it currently seems to spawn a new webpack for every refresh, consuming my system memory?

turbocrime: webpack should have a way to output a pid, or a way to instruct it to reload modules. and if that isn't feasible it should be wrapped in something that can. pkill is the wrong approach

watch-and-repack.sh
TalDerei: ah good point, modified to echo an error message and exit if unset.

grod220 (dismissed):

turbocrime (approved):
still think there is a better way to acquire webpack pid but approving

PR #53, opened 2024-06-24
fix(extension): #37: add loader to rpc validation

### LINKED ISSUE #37: RPC validation indicator
During onboarding there is a step to select an RPC. After making a choice, it will query for its chainId. After a successful response, it shows this ID under the `Next` button (which is now clickable):
 
Screenshot 2024-06-17 at 2 02 30 PM

This serves as a kind of rpc validation step. If the user selects one that isn't valid they get notified:

Screenshot 2024-06-17 at 2 04 29 PM

However, if the RPC is slow, this query could take some time. Unfortunately, we aren't given an indication of what's happening during this step:

Screenshot 2024-06-17 at 2 02 01 PM

This has caused some users to be confused why they cannot proceed. We should have a loading indicator that informs the user that a fetch is currently happening to validate their entry.

VanishMax:
Closes #37 

Add a loader to the RPC selection setting page

image

### DIFF
--- a/apps/extension/src/shared/components/grpc-endpoint-form/index.tsx
+++ b/apps/extension/src/shared/components/grpc-endpoint-form/index.tsx
@@ -1,7 +1,7 @@
 import { FormEvent, useRef } from 'react';
 import { SelectList } from '@repo/ui/components/ui/select-list';
 import { Button } from '@repo/ui/components/ui/button';
-import { Network } from 'lucide-react';
+import { Network, Loader2 } from 'lucide-react';
 import { useGrpcEndpointForm } from './use-grpc-endpoint-form';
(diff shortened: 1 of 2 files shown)


apps/extension/src/shared/components/grpc-endpoint-form/index.tsx
```diff
@@ -98,7 +99,7 @@ export const GrpcEndpointForm = ({
           </div>
 
           <Button variant='gradient' type='submit' disabled={!isSubmitButtonEnabled}>
-            {submitButtonLabel}
+            {isValidationLoading ? <Loader2 className='size-4 animate-spin' /> : submitButtonLabel}
```
grod220, line 102: Could we add text to go along with the loading spinner? Something like: `Validating rpc <spinner>`

grod220 (approved):

PR #56, opened 2024-06-25
fulfill disconnect interface

turbocrime:
fulfill the disconnect interface published in`@penumbra-zone/client@9`


apps/extension/src/senders/approve.ts
turbocrime, line 1: some mild reorganization

apps/extension/src/storage/origin.ts
turbocrime, line 1: storage utilities moved to storage dir

apps/extension/src/storage/types.ts
```diff
@@ -22,6 +22,7 @@ export interface LocalStorageState {
   frontendUrl: string | undefined;
   passwordKeyPrint: KeyPrintJson | undefined;
   fullSyncHeight: number | undefined;
+  // TODO: It's likely that an array is not the best data structure for this in storage. Should revisit later.
```
turbocrime, line 25: moved this comment to the storage types definition

apps/extension/src/listeners/message-prax-disconnect.ts
turbocrime, line 1: the new handler

apps/extension/src/senders/disconnect.ts
turbocrime, line 1: disconnect implementation

apps/extension/public/manifest.json
```diff
@@ -41,7 +40,7 @@
   },
   "web_accessible_resources": [
     {
-      "resources": ["manifest.json"],
+      "resources": ["manifest.json", "favicon/*"],
```
turbocrime, line 43: expose the extension icon so dapps may use it

apps/extension/src/content-scripts/injected-penumbra-global.ts
turbocrime, line 1: injection is refactored into a class to more clearly and explicitly manage state.

the interface is async. so to provide the synchronous `isConnected` state querying method, the state of the three possible actions are tracked, and resolved in the private function `state`.

as expected, `isConnected() => true` would indicate an active connection.

more information on the interface is available in client package documentation, but what it comes down to is:

- if the page has called `disconnect`, the connection state is false.
- if the page has called `request` but the request failed, the connection state is false.
- if no action has been attempted, the connection state is undefined.
- if connection has succeeded, the connection state is true.

grod220 (approved):

grod220:
Going to take this PR over, rebase, and merge so we can cut a release given the proving key update

turbocrime:
thanks

PR #58, opened 2024-06-25
fix: trim trailing slash from custom rpc url

vacekj:
Fixes edge case where entering a custom RPC URL with a trailing slash will result in links in other parts of the app being broken.

@grod220 I couldn't find a corresponding test file for this form - should I create one?


vacekj:
Agree.

the input is already marked as `type="url"`, so that provides enough validation imo. In terms of joining the URL, we can indeed use the URL constructor:

```javascript
new URL('/one', 'http://example.com/').href
```

which works with or without a trailing slash.

vacekj:
@turbocrime requesting re-review, lmk if there's anything else I should change.

turbocrime (approved):
looks good to me

PR #61, opened 2024-06-26
Guard against bad default frontends

grod220:
Scopes default frontends to http & https only for security purposes

### DIFF
--- a/apps/extension/src/shared/utils/is-valid-url.ts
+++ b/apps/extension/src/shared/utils/is-valid-url.ts
@@ -1,6 +1,13 @@
 export const isValidUrl = (url: string) => {
   try {
-    new URL(url);
+    const proposedUrl = new URL(url);
+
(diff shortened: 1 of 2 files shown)


apps/extension/src/shared/utils/is-valid-url.ts
```diff
@@ -1,6 +1,13 @@
 export const isValidUrl = (url: string) => {
   try {
-    new URL(url);
+    const proposedUrl = new URL(url);
+
+    // Security measure: allows us to guard against someone being deceived
+    // into adding a url with chrome-extension:// in it with intentions to inject code
+    if (!['https:', 'http:'].includes(proposedUrl.protocol)) {
+      throw new Error(`protocol "${proposedUrl.protocol}" is not supported`);
```
VanishMax, line 8: nitpick: `return false` could be clearer in here

VanishMax (approved):

PR #64, opened 2024-06-26
fix(extension): #32: save chainIn in localstorage

### LINKED ISSUE #32: 'params' are never stored in local storage
found at https://github.com/prax-wallet/web/pull/28#discussion_r1638084606

https://github.com/prax-wallet/web/blob/236e6502a88ca6f5e8e030a0c2a14b9ed521f340/apps/extension/src/wallet-services.ts#L39-L41

We retrieve 'params' from local storage to get the `chain-id` without having to make any network requests, but in fact the 'params' are always empty because we never save them to local storage

related to https://github.com/penumbra-zone/web/pull/998

VanishMax:
Closes #32 

Stores `chainId` in chrome extension storage on RPC change, so it doesn't require a request from the service worker later.

Decided to store `chainId` instead of `params` since the only used property from AppParameters is `chainId`


Valentine1898:
The thing is that `chain_id` can change not only when the user changes the RPC

VanishMax:
Closing as it is duplicated by #65

PR #65, opened 2024-06-26
save app params

### LINKED ISSUE #32: 'params' are never stored in local storage

turbocrime:
at some point saving app params to local was removed.

this restores ability to operate offline.

edit: fixes #32 
accidentally duplicating work in #64

### DIFF
--- a/.changeset/green-frogs-shake.md
+++ b/.changeset/green-frogs-shake.md
@@ -1,0 +1,5 @@
+---
+'chrome-extension': patch
+---
+
+save app params
--- a/apps/extension/src/wallet-services.ts
+++ b/apps/extension/src/wallet-services.ts
@@ -38,19 +38,17 @@ export const startWalletServices = async () => {
  * local storage.
  */
 const getChainId = async (baseUrl: string) => {
-  const localChainId = await localExtStorage
-    .get('params')
-    .then(json => json && AppParameters.fromJsonString(json).chainId);
+  const params =
+    (await createPromiseClient(AppService, createGrpcWebTransport({ baseUrl })).appParameters({}))
+      .appParameters ??
+    (await localExtStorage
+      .get('params')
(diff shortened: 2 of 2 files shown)


Valentine1898:
It seems that this solution, as well as #64, does not handle the case where the chain-id can change without changing the RPC

turbocrime:
chain-id should not change for the lifetime of the chain. if chain-id changes without changing the rpc, the rpc is serving a new chain. at this point all local state is invalid.

since chain-id is part of idb name, services will connect to a (probably new) idb database correct for the chain, and begin sync. this is equivalent to clearing the cache, but the old chain-id state remains the old database name.

Valentine1898:
All the time we had only one RPC for the testnet, while the chain-id changed many times. If RPC starts serving a new chain, then when indexed-db is initialized, the chain-id from local storage (old chain) will be used. This means that the data of the new chain will be written to the same database on top of the old data, and this will probably lead to various bugs

turbocrime:
you're correct, sorry. i had some logic previously in sercices-context init that i believe was comprehensive

turbocrime:
on phone rn and difficult to link properly but ab9d743f8ab845f746ac31418ba1a91578d4a928 of penumbra-zone web should contain something usable that could be copied from services-context index

Valentine1898:
We can actually prioritize remote params, and only if they are not available use locally stored params

Valentine1898 (approved):
LGTM!
But it seems that extension without network doesn't work as expected (you can't see your balances)

PR #85, opened 2024-07-02
improve fee rendering in approval dialogue 

TalDerei:
references https://github.com/penumbra-zone/web/issues/1311 and https://github.com/penumbra-zone/web/issues/1358. One of the side-effects of the migration is that these same changes should be propagated to the web ui package for minifront to consume. 

- fees don’t have metadata, so modifies swap claim fees to display the proper amounts and denoms (rather than incorrectly hardcoding **upenumbra** for all trades) 
- fixes pill size in transaction approval dialogue 

--------------------------------------------------------------------------------------------------------------------------

**before**

Screenshot 2024-07-04 at 12 56 35 PM

--------------------------------------------------------------------------------------------------------------------------
**after**

Screenshot 2024-07-04 at 12 55 42 PM


grod220 (approved):
Left comments in https://github.com/penumbra-zone/web/pull/1419

PR #86, opened 2024-07-03
webpack watch tarballs

turbocrime:
improves local tarball workflow.

run

```sh
pnpm -w add /absolute/path/to/your/repo/*/namespace-tarballs-*.tgz
pnpm -w syncpack fix-mismatches
pnpm dev
```

in this repository to use local tarballs, and automatically reload changes when those tarballs update.


TalDerei (changes requested):
can you please restore [watch-and-repack.sh](https://github.com/prax-wallet/web/pull/86/files#diff-c9dfcbf51eda9b1f9296dc4de53a7e711c741f199bd63ebf8d3129e5e97a27a0)? I've found it extremely useful for my workflow, and there's detailed instructions around using it. this can be an additional way to do local dev and keeping it doesn't hurt 

let's also move it to the scripts folder so it's not in the root

apps/extension/webpack.config.ts
```diff
@@ -147,10 +190,11 @@ const config: webpack.Configuration = {
       filename: 'offscreen.html',
       chunks: ['offscreen-handler'],
     }),
+    // watch tarballs for changes
+    WEBPACK_WATCH && new WatchExternalFilesPlugin({ files: localPackages }),
+    WEBPACK_WATCH && PnpmInstallPlugin,
```
TalDerei, line 195: was aware of built-in hot module reloading for webpack config (https://webpack.js.org/configuration/watch/), but wasn't aware of the watch options (https://webpack.js.org/configuration/watch/).

TalDerei (approved):
confirmed offline that script properly propagates changes. 

general curiosity, what prevents it from falling into an infinite loop? when I simulated hmr with `fswatch`, it initially continuously triggered the repack process because the repack itself was causing changes that fswatch picked up, leading to an infinite loop.

turbocrime:
> general curiosity, what prevents it from falling into an infinite loop? when I simulated hmr with `fswatch`, it initially continuously triggered the repack process because the repack itself was causing changes that fswatch picked up, leading to an infinite loop.

You may have been `fswatch`ing files you were changing. This configuration only watches the tarballs, and it doesn't trigger any activity in the watched locations.

apps/prax-marketing-site/vite.config.ts
```diff
@@ -2,6 +2,7 @@ import { defineConfig } from 'vite';
 import react from '@vitejs/plugin-react-swc';
 
 export default defineConfig({
+  clearScreen: false,
```
grod220, line 5: what does this do?

packages/tsconfig/base.json
```diff
@@ -24,7 +23,7 @@
     "noUnusedParameters": true,
     "resolveJsonModule": true,
     "target": "ESNext",
-    "module": "ES2022"
+    "module": "preserve"
```
grod220, line 26: what are the implications of this new module config?

apps/prax-marketing-site/vite.config.ts
turbocrime: this prevents the vite build from blanking the terminal, so extension build output remains visible when running `dev` from the repository root

packages/tsconfig/base.json
turbocrime: there are zero implications because no packages are published from this repo. but due to several idiosyncratic configurations and webpack's lack of true support for typescript config, the previous module resolution prevented the import of the package json data in webpack config, and this resolution mode is more permissive.

PR #87, opened 2024-07-03
extension reloading webpack plugin

turbocrime:
this adds a webpack plugin that

1. creates a local profile
2. provides extension HMR

content scripts can't re-inject, so you'll still have to reload pages.

you can activate the plugin by specifying a chromium profile directory, either outside the repository or at the conveniently gitignored 'chromium-profile'

```sh
CHROMIUM_PROFILE='chromium-profile' pnpm dev
```


apps/extension/web-ext.d.ts
turbocrime, line 1: web-ext doesn't have types, so this is ripped verbatim from an issue on the web-ext repo

apps/extension/webpack.config.ts
```diff
@@ -39,6 +45,27 @@ const DefinePlugin = new webpack.DefinePlugin({
...
+          profileCreateIfMissing: Boolean(CHROMIUM_PROFILE),
+          sourceDir: options.output.path,
+          startUrl: 'https://localhost:5173/',
+        });
+        this.webExtRun.registerCleanup(() => (this.webExtRun = undefined));
+      },
+    );
+  },
+};
+
```
turbocrime, line 68: this is much like the small webpack plugin in the initial transport-demo

apps/extension/web-ext.d.ts
turbocrime: smaller type def now

apps/extension/webpack.config.ts
```diff
@@ -39,6 +45,27 @@ const DefinePlugin = new webpack.DefinePlugin({
...
 
+const WebExtReloadPlugin = {
+  webExtRun: undefined as WebExtRunner | undefined,
+  apply({ hooks }: webpack.Compiler) {
+    hooks.afterEmit.tapPromise(
+      { name: 'WebExt Reloader' },
+      async ({ options }: webpack.Compilation) => {
+        await this.webExtRun?.reloadAllExtensions();
+        this.webExtRun ??= await WebExtCmd.run({
+          target: 'chromium',
```
turbocrime, line 56: target is labelled 'chromium' but it will execute actual Chrome by default

TalDerei (approved):

PR #92, opened 2024-07-05
fulfill injection state interfaces

turbocrime:
this PR fulfills the interface defined in https://github.com/penumbra-zone/web/pull/1420

published versions of the client package now expect this interface to be fulfilled

the react wrapper also depends on this interface

### DIFF
--- a/apps/extension/src/content-scripts/injected-penumbra-global.ts
+++ b/apps/extension/src/content-scripts/injected-penumbra-global.ts
@@ -4,7 +4,7 @@
  * to them.
  *
  * The global is identified by `Symbol.for('penumbra')` and consists of a record
- * with string keys referring to `PenumbraInjection` objects that contain a
+ * with string keys referring to `PenumbraProvider` objects that contain a
  * simple API. The identifiers on this record should be unique, and correspond
  * to a browser extension id. Providers should provide a link to their extension
  * manifest in their record entry.
@@ -17,16 +17,16 @@
  * connections.
  */
 
-import { PenumbraInjection, PenumbraSymbol } from '@penumbra-zone/client';
+import { PenumbraProvider, PenumbraState, PenumbraSymbol } from '@penumbra-zone/client';
+import { PenumbraStateEvent } from '@penumbra-zone/client/event';
 
+import { PraxConnection } from '../message/prax';
 import {
   isPraxFailureMessageEvent,
   isPraxPortMessageEvent,
   PraxMessage,
   unwrapPraxMessageEvent,
 } from './message-event';
-
-import { PraxConnection } from '../message/prax';
 
 type PromiseSettledResultStatus = PromiseSettledResult<unknown>['status'];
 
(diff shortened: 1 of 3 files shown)


turbocrime:
cc @VanishMax

apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -46,63 +51,87 @@ class PraxInjection {
...
+
+    state: () => this.reduceInjectionState(),
+
+    manifest: String(this.manifestUrl),
+
+    addEventListener: ((...params) =>
+      this.stateEvents.addEventListener(...params)) as EventTarget['addEventListener'],
+
+    removeEventListener: ((...params) =>
+      this.stateEvents.removeEventListener(...params)) as EventTarget['removeEventListener'],
```
VanishMax, line 88: suggestion: instead of emitting events, simply provide callback `onStateChange` for simpler DX.

I tested the interfaces of the `client` package and it feels that custom events require more understanding from users. I personally had problems with typing the event handler in the test application. It is a problem that either requires users to search for workarounds or for us to publish more helper types. And it's just hard...

A simple callback is fine in our case and we don't necessarily need to allow removing listeners.

Alternative: rename them to `addListener` and `removeListener` and rename the event name to simply `state`, return plain object with `origin` and `state` fields instead of CustomEvent. This way we keep the subscribe-unsubscribe logic but, in my opinion, it's not really important.

VanishMax: yeah, i reviewed the the mentioned PR and now struggling to see the further development. I have a feeling that we should postpone the merge of this and the react package (https://github.com/penumbra-zone/web/pull/1450) before the strategic alignment on how the injections should look like

turbocrime: cc @grod220 you asked for a callback style interface in https://github.com/penumbra-zone/web/pull/1293#discussion_r1660844191

that interface was reviewed and merged, this PR is attempting to actually implement that interface for prax

trying to move forward with https://github.com/penumbra-zone/web/pull/1450 which, as suggested, wraps the injection using the callback interface to simplify state management

can we either 

1. merge this as meeting that request
2. merge this as progress towards that request
3. identify a clear reason this doesn't meet that request
4. elaborate an actual spec

turbocrime: > I personally had problems with typing the event handler in the test application. It is a problem that either requires users to search for workarounds or for us to publish more helper types. And it's just hard...

pr now updated to more specifically type the addEventListener and removeEventListener methods to only accept attaching listeners to the specific event type exported, in an extensible way.
 
> Alternative: rename them to `addListener` and `removeListener` and rename the event name to simply `state`, return plain object with `origin` and `state` fields instead of CustomEvent.

the current interface extends the standard DOM EventTarget interface, and CustomEvent is the standard way to use custom events at an EventTarget. using a very generic event name is probably a bad idea.

https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/EventTarget

VanishMax (approved):

PR #96, opened 2024-07-05
Fix script permissions

jessepinho:
Without this change, running 'pnpm watch-and-repack' fails due to missing execution permissions.


turbocrime (approved):
approving but you should use the webpack plugin for reloading now described in https://github.com/prax-wallet/web/pull/86

also check out https://github.com/prax-wallet/web/pull/91

PR #117, opened 2024-07-19
fix: don't auto-animate children of box

### LINKED ISSUE #78: Animation height squeeze
Upon toggling the public/private view of a transaction, the transition has a strange vertical resizing animation. Think this is likely because we need to wrap this component in the Framer motion div. 

https://github.com/prax-wallet/web/assets/16624263/70969bdc-3881-401f-b86a-3b604e823c03

vacekj:
WIP: waiting on @jessepinho to confirm the parent-box layout animation doesn't break anything.

fixes #78

### DIFF
--- a/packages/ui/components/ui/box/index.tsx
+++ b/packages/ui/components/ui/box/index.tsx
@@ -74,11 +74,7 @@ export const Box = ({
         </div>
       )}
 
-      {children && (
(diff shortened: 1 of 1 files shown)


jessepinho (approved):
Huh, I'm surprised this fixes the issue. Can you add back the `layout` prop to the `motion.div`? Otherwise, should be fine — we'll be replacing this component soon anyway.

PR #125, opened 2024-07-24
feat(extension): display images for registry frontends

VanishMax:
Relates to https://github.com/prax-wallet/registry/issues/60

image

### DIFF
--- a/apps/extension/src/shared/components/default-frontend-form/index.tsx
+++ b/apps/extension/src/shared/components/default-frontend-form/index.tsx
@@ -20,7 +20,7 @@ const useFrontendsList = (selectedRpc?: string) => {
         ...arr,
         {
           name: 'Embedded RPC frontend',
-          /*NB: we merge using the variadic URL constructor here to avoid double-slashes*/
+          /* we merge using the variadic URL constructor here to avoid double-slashes*/
           url: new URL('/app/', selectedRpc).href,
           images: [],
         },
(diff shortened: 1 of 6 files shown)


apps/extension/src/shared/components/default-frontend-form/index.tsx
```diff
@@ -1,39 +1,39 @@
...
-  }));
+  const frontends = useMemo(() => {
+    const arr: EntityMetadata[] = data?.frontends ?? [];
+
+    if (selectedRpc) {
+      return [
+        ...arr,
+        {
+          name: 'Embedded RPC frontend',
+          /*NB: we merge using the variadic URL constructor here to avoid double-slashes*/
```
grod220, line 23: question: what's `NB` mean? 🤔

apps/extension/src/shared/components/default-frontend-form/index.tsx
```diff
@@ -58,16 +58,28 @@ export const DefaultFrontendForm = ({ isOnboarding }: { isOnboarding?: boolean }
...
-          key={option.url}
-          value={option.url}
-          secondary={option.url}
-          label={option.title}
-          isSelected={option.url === selectedFrontend}
-          onSelect={selectUrl}
-        />
-      ))}
+      {frontends.map(option => {
+        const imageUrl = option.images[0]?.png ?? option.images[0]?.svg;
```
grod220, line 62: suggestion: we should probably prefer svg right?

apps/extension/src/shared/components/default-frontend-form/index.tsx
VanishMax: no idea, this comment was there before me 🙈

PR #129, opened 2024-07-26
fix(extension): #107: allow http for localhost

### LINKED ISSUE #107: Accept localhost connections without https
External teams are running into issues in local dev connecting to Prax because it has a strict https-only requirement. By default localhost is generally accessed through http. Changing web build tooling to point to https locally is a bit clunky. 

**Proposal:** allow Prax to make connections in a dev environment to http://localhost

This will circumvent the ongoing issue of dev teams reaching out to us because of mysterious connection issues.

VanishMax:
Closes #107 

1. Changes to the manifest to inject content scripts into http://localhost
2. Changes to the `assertValidSender` function to allow http://localhost

### DIFF
--- a/apps/extension/public/manifest.json
+++ b/apps/extension/public/manifest.json
@@ -16,7 +16,7 @@
   },
   "content_scripts": [
     {
-      "matches": ["https://*/*"],
+      "matches": ["https://*/*", "http://localhost:*/*"],
       "js": [
         "injected-connection-port.js",
         "injected-disconnect-listener.js",
@@ -25,7 +25,7 @@
       "run_at": "document_start"
     },
     {
-      "matches": ["https://*/*"],
+      "matches": ["https://*/*", "http://localhost:*/*"],
       "js": ["injected-penumbra-global.js"],
       "run_at": "document_start",
       "world": "MAIN"
--- a/apps/extension/src/senders/validate.ts
+++ b/apps/extension/src/senders/validate.ts
@@ -6,16 +6,12 @@ type ValidSender = chrome.runtime.MessageSender & {
   frameId: 0;
   documentId: string;
   tab: chrome.tabs.Tab & { id: number };
-
-  // the relationship between origin and url is pretty complex.
-  // just rely on the browser's tools.
-  origin: `${ValidProtocol}//${string}`;
-  url: `${ValidProtocol}//${string}/${string}`;
+  origin: string;
+  url: string;
 };
 
-const isException = (url: URL): boolean => {
-  return url.protocol === 'http:' && url.hostname === 'localhost';
-};
+const isHttpLocalhost = (url: URL): boolean =>
+  url.protocol === 'http:' && url.hostname === 'localhost';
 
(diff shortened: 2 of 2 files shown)


apps/extension/src/senders/validate.ts
```diff
@@ -13,6 +13,10 @@ type ValidSender = chrome.runtime.MessageSender & {
   url: `${ValidProtocol}//${string}/${string}`;
 };
 
+const isException = (url: URL): boolean => {
+  return url.protocol === 'http:' && url.hostname === 'localhost';
+};
+
```
turbocrime, line 19:
```suggestion
const isHttpLocalhost = (url: URL): boolean => 
  url.protocol === 'http:' && url.hostname === 'localhost';

```

apps/extension/src/senders/validate.ts
```diff
@@ -34,7 +38,8 @@ export const assertValidSender = (sender?: chrome.runtime.MessageSender) => {
   if (parsedOrigin.origin !== sender.origin) {
     throw new Error('Sender origin is invalid');
   }
-  if (!(parsedOrigin.protocol in ValidProtocol)) {
+
+  if (!(parsedOrigin.protocol in ValidProtocol || isException(parsedOrigin))) {
```
turbocrime, line 42:
```suggestion
  if (!(parsedOrigin.protocol in ValidProtocol || (globalThis.__DEV__ && isHttpLocalhost(parsedOrigin)))) {
```

turbocrime (approved):

apps/extension/public/manifest.json
turbocrime, line 1: i'd like to condition the manifest inclusion of `localhost` but that would require some additional build config

apps/extension/src/senders/validate.ts
```diff
@@ -13,6 +13,10 @@ type ValidSender = chrome.runtime.MessageSender & {
   url: `${ValidProtocol}//${string}/${string}`;
```
turbocrime, line 13: the origin/url type are incorrect now that validprotocol may conditionally be http at any moment. probably just set these to 'string'

turbocrime, line 13: the origin/url type are incorrect now that validprotocol may conditionally be http at any moment. probably just set these to 'string'

turbocrime (changes requested):
the ValidSender type needs correction but other than that, this would work

turbocrime:
deps also need to support this https://github.com/penumbra-zone/web/blob/9da630e851cd7303372b3e1aac63d398305bba62/packages/transport-chrome/src/session-manager.ts#L99-L105

apps/extension/public/manifest.json
VanishMax: what do you mean by this?

apps/extension/src/senders/validate.ts
VanishMax: changed to `string`

VanishMax:
@turbocrime pinging for your re-review

PR #153, opened 2024-08-12
fix(extension): #107: allow http for localhost

### LINKED ISSUE #107: Accept localhost connections without https

VanishMax:
Closes #107 

Relates to #129 and the reverting PR #145.

Allows http protocol for localhost


apps/extension/public/manifest.json
```diff
@@ -16,12 +16,12 @@
   },
   "content_scripts": [
     {
-      "matches": ["https://*/*"],
+      "matches": ["https://*/*", "http://localhost/*"],
```
VanishMax, line 19: this line, instead of the previously used one, must work correctly according to the docs: https://developer.chrome.com/docs/extensions/develop/concepts/match-patterns#special

grod220: Just checked a test upload, it works 👍

apps/extension/src/senders/validate.ts
```diff
@@ -34,7 +34,13 @@ export const assertValidSender = (sender?: chrome.runtime.MessageSender) => {
...
     throw new Error('Sender origin is invalid');
   }
-  if (!(parsedOrigin.protocol in ValidProtocol)) {
+
+  if (
+    !(
+      parsedOrigin.protocol in ValidProtocol ||
+      (globalThis.__DEV__ && isHttpLocalhost(parsedOrigin))
+    )
+  ) {
```
grod220, line 43: todo: we should have a test for this in validate.test.ts. Also, could you check that test suite to ensure we've covered all cases for this function?

VanishMax: @grod220 thanks for pointing this out. I should pay more attention to the tests

grod220 (approved):

apps/extension/src/senders/validate.ts
```diff
@@ -34,7 +34,13 @@ export const assertValidSender = (sender?: chrome.runtime.MessageSender) => {
   if (parsedOrigin.origin !== sender.origin) {
     throw new Error('Sender origin is invalid');
   }
-  if (!(parsedOrigin.protocol in ValidProtocol)) {
+
+  if (
+    !(
+      parsedOrigin.protocol in ValidProtocol ||
+      (globalThis.__DEV__ && isHttpLocalhost(parsedOrigin))
```
grod220, line 41: Stumbled upon this today. I was trying to connect a production Prax to the dev Dex explorer and it would silently fail. Think we should drop the requirement that the dev version of Prax is needed to connect to localhost sites:

```diff
parsedOrigin.protocol in ValidProtocol ||
- (globalThis.__DEV__ && isHttpLocalhost(parsedOrigin))
+ isHttpLocalhost(parsedOrigin)
```

@VanishMax can you open a follow PR on this?

VanishMax: @grod220 no problem! https://github.com/prax-wallet/prax/pull/208

PR #164, opened 2024-08-19
sync: fresh and existing wallets skip trial decryption 

TalDerei:
references https://github.com/penumbra-zone/web/issues/1278 and https://github.com/penumbra-zone/web/pull/1707

Implements the necessary scaffolding for wallet birthdays in prax


apps/extension/src/routes/page/onboarding/generate.tsx
```diff
@@ -14,20 +14,74 @@ import { generateSelector } from '../../../state/seed-phrase/generate';
...
+import { localExtStorage } from '../../../storage/local';
+import { sample } from 'lodash';
+
+const fetchBlockHeightWithFallback = async (endpoints: string[]): Promise<number> => {
+  if (endpoints.length === 0) {
+    throw new Error('All RPC endpoints failed to fetch the block height.');
+  }
+
+  // Randomly select an rpc endpoint from the chain registry
+  const randomEndpoint = sample(endpoints);
```
TalDerei, line 28: lifting from [migration](https://github.com/prax-wallet/web/blob/main/apps/extension/src/storage/migrations/local-v1-migration.ts#L100) function

apps/extension/src/routes/page/onboarding/generate.tsx
```diff
@@ -14,20 +14,74 @@ import { generateSelector } from '../../../state/seed-phrase/generate';
...
+    } else {
+      // Remove the current endpoint from the list and try again
+      const remainingEndpoints = endpoints.filter(endpoint => endpoint !== randomEndpoint);
+      return fetchBlockHeightWithFallback(remainingEndpoints);
+    }
+  } catch (error) {
+    const remainingEndpoints = endpoints.filter(endpoint => endpoint !== randomEndpoint);
+    return fetchBlockHeightWithFallback(remainingEndpoints);
+  }
+};
```
TalDerei, line 43: this fetches the block height by randomly querying one of the rpc endpoints in the chain registry. something to consider is whether this operation is possibly blocking or if the rpc call will eventually time out?

apps/extension/src/routes/page/onboarding/generate.tsx
```diff
@@ -60,6 +114,21 @@ export const GenerateSeedPhrase = () => {
...
+                Block Height:{' '}
+                <span className='font-bold text-gray-100'>{blockHeight ?? 'Loading...'}</span>
+              </p>
+              <p className='mt-2 text-sm text-gray-400'>
+                Please save the wallet creation height along with your recovery passphrase. It will
+                help you restore your wallet more quickly!
+              </p>
+            </div>
+          )}
+
```
TalDerei, line 131: Screenshot 2024-08-28 at 10 42 09 PM

apps/extension/src/routes/page/onboarding/height.tsx
```diff
@@ -0,0 +1,73 @@
...
+          <CardTitle className='font-semibold'>
+            Enter Block Height When the Wallet Was Created (Optional)
+          </CardTitle>
+          <CardDescription>
+            This step is optional. Providing your wallet&apos;s block creation height can help speed
+            up the synchronization process, but it&apos;s not required. If you don&apos;t have this
+            information, you can safely skip this step.
+          </CardDescription>
+        </CardHeader>
+        <CardContent>
```
TalDerei, line 56: Screenshot 2024-08-28 at 10 42 48 PM

apps/extension/src/hooks/full-sync-height.ts
```diff
@@ -55,3 +55,14 @@ export const useSyncProgress = () => {
...
+export const fetchBlockHeight = async (grpcEndpoint: string) => {
+  const tendermintClient = createPromiseClient(
+    TendermintProxyService,
+    createGrpcWebTransport({ baseUrl: grpcEndpoint }),
+  );
+  const blockHeight = (await tendermintClient.getStatus({}).catch(() => undefined))?.syncInfo
+    ?.latestBlockHeight;
+
+  return blockHeight ? Number(blockHeight) : undefined;
+};
```
grod220, line 68: issue: this seems to be redundant to the function nested in the hook above. However, looks like this one tries to return an `undefined` in a failure case (don't think that's necessary given the consumer uses a try/catch). Can we extract the original and use it for both places?

apps/extension/src/routes/page/onboarding/generate.tsx
```diff
@@ -14,20 +14,74 @@ import { generateSelector } from '../../../state/seed-phrase/generate';
 import { usePageNav } from '../../../utils/navigate';
 import { PagePath } from '../paths';
 import { WordLengthToogles } from '../../../shared/containers/word-length-toogles';
+import { ChainRegistryClient } from '@penumbra-labs/registry';
+import { fetchBlockHeight } from '../../../hooks/full-sync-height';
+import { localExtStorage } from '../../../storage/local';
+import { sample } from 'lodash';
+
+const fetchBlockHeightWithFallback = async (endpoints: string[]): Promise<number> => {
```
grod220, line 22: todo: let's add a code comment on what this function does above

apps/extension/src/routes/page/onboarding/generate.tsx
```diff
@@ -14,20 +14,74 @@ import { generateSelector } from '../../../state/seed-phrase/generate';
...
+        const blockHeight = await fetchBlockHeightWithFallback(suggestedEndpoints);
+        await localExtStorage.set('walletCreationBlockHeight', blockHeight);
+        setBlockHeight(blockHeight);
+
+        isFetchedRef.current = true;
+      } catch (error) {
+        console.error('Error fetching data:', error);
+      }
+    };
+
```
grod220, line 76: issue: think we should move all this logic to the SeedPhraseSlice or somewhere in zustand (fetchBlockHeightWithFallback too)

apps/extension/src/routes/page/onboarding/generate.tsx
grod220: I think we added timeouts to our promise clients (?). It's async, so not thread blocking, but we should have a think whether this should block the onboarding step.

apps/extension/src/routes/page/onboarding/generate.tsx
```diff
@@ -14,20 +14,74 @@ import { generateSelector } from '../../../state/seed-phrase/generate';
...
 import { WordLengthToogles } from '../../../shared/containers/word-length-toogles';
+import { ChainRegistryClient } from '@penumbra-labs/registry';
+import { fetchBlockHeight } from '../../../hooks/full-sync-height';
+import { localExtStorage } from '../../../storage/local';
+import { sample } from 'lodash';
+
+const fetchBlockHeightWithFallback = async (endpoints: string[]): Promise<number> => {
+  if (endpoints.length === 0) {
+    throw new Error('All RPC endpoints failed to fetch the block height.');
+  }
```
grod220, line 25: At the moment, think this error will bubble up. We should move this to zustand, capture error states, and display them in the UI.

apps/extension/src/routes/page/onboarding/generate.tsx
```diff
@@ -14,20 +14,74 @@ import { generateSelector } from '../../../state/seed-phrase/generate';
...
+
+  // Randomly select an rpc endpoint from the chain registry
+  const randomEndpoint = sample(endpoints);
+
+  try {
+    const walletCreationBlockHeight = await fetchBlockHeight(randomEndpoint!);
```
grod220, line 31: issue: In the other repo, we now have an eslint rule for us to avoid using `!`. Can you have an error guard above instead?

```ts
  if (!randomEndpoint) {
    throw new Error('No RPC endpoints found.');
  }
```

apps/extension/src/routes/page/onboarding/generate.tsx
```diff
@@ -14,20 +14,74 @@ import { generateSelector } from '../../../state/seed-phrase/generate';
...
+    if (walletCreationBlockHeight !== undefined) {
+      return walletCreationBlockHeight;
```
grod220, line 33: issue: This is related to my other comment. If there is an error, it will go to the catch clause. Don't think we need to return undefined from the function. This will also prevent the need to have the double `const remainingEndpoints...` lines.

apps/extension/src/routes/page/onboarding/generate.tsx
```diff
@@ -60,6 +114,21 @@ export const GenerateSeedPhrase = () => {
...
+          {reveal && (
+            <div className='mt-4 rounded-lg border border-gray-500 bg-gray-800 p-4 shadow-sm'>
+              <h4 className='text-lg font-semibold text-gray-200'>Important!</h4>
+              <p className='mt-2 text-gray-300'>
+                Block Height:{' '}
+                <span className='font-bold text-gray-100'>{blockHeight ?? 'Loading...'}</span>
+              </p>
+              <p className='mt-2 text-sm text-gray-400'>
+                Please save the wallet creation height along with your recovery passphrase. It will
+                help you restore your wallet more quickly!
```
grod220, line 127: suggestion:

````diff
- It will help you restore your wallet more quickly!
+ It's not required, but will help you restore your wallet quicker on a fresh Prax install next time.
```

apps/extension/src/routes/page/onboarding/generate.tsx
```diff
@@ -60,6 +114,21 @@ export const GenerateSeedPhrase = () => {
               isSuccessCopyText
             />
           </div>
+
+          {reveal && (
+            <div className='mt-4 rounded-lg border border-gray-500 bg-gray-800 p-4 shadow-sm'>
+              <h4 className='text-lg font-semibold text-gray-200'>Important!</h4>
```
grod220, line 120: suggestion: Think we can get rid of this headline and graduate block height line to an `h4`.

apps/extension/src/routes/page/onboarding/height.tsx
```diff
@@ -0,0 +1,73 @@
...
+  return (
+    <FadeTransition>
+      <BackIcon className='float-left mb-4' onClick={() => navigate(-1)} />
+      <Card className='w-[600px] p-6' gradient>
+        <CardHeader className='items-center'>
+          <CardTitle className='font-semibold'>
+            Enter Block Height When the Wallet Was Created (Optional)
+          </CardTitle>
+          <CardDescription>
+            This step is optional. Providing your wallet&apos;s block creation height can help speed
```
grod220, line 51: suggestion: Think we can axe `This step is optional.` It's mentioned in the headline and says it's not required below.

apps/extension/src/routes/page/onboarding/height.tsx
```diff
@@ -0,0 +1,73 @@
...
+    });
+  };
+
+  return (
+    <FadeTransition>
+      <BackIcon className='float-left mb-4' onClick={() => navigate(-1)} />
+      <Card className='w-[600px] p-6' gradient>
+        <CardHeader className='items-center'>
+          <CardTitle className='font-semibold'>
+            Enter Block Height When the Wallet Was Created (Optional)
```
grod220, line 48: question: is this wording similar to zcash's? Would it be more accurate to say "Enter the block height of your first Penumbra transaction?". Not sure.

suggestion: Let's just capitalize the first word

apps/extension/src/routes/page/onboarding/height.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const navigate = usePageNav();
+  const [blockHeight, setBlockHeight] = useState('');
+
+  const handleSubmit = async (event: MouseEvent | FormEvent) => {
+    event.preventDefault();
+
+    if (blockHeight) {
+      // Save the block height to local extension storage
+      await localExtStorage.set('walletCreationBlockHeight', Number(blockHeight));
+    }
```
grod220, line 27: question: Local storage setting seems to happen at the very end of onboarding right? See `useOnboardingSave()` hook. If you use zustand, you can just add logic there to save `walletCreationBlockHeight`

apps/extension/src/routes/page/onboarding/height.tsx
```diff
@@ -0,0 +1,73 @@
...
+  CardTitle,
+} from '@repo/ui/components/ui/card';
+import { FadeTransition } from '@repo/ui/components/ui/fade-transition';
+import { usePageNav } from '../../../utils/navigate';
+import { PagePath } from '../paths';
+import { FormEvent, MouseEvent, useState } from 'react';
+import { Input } from '@repo/ui/components/ui/input';
+import { localExtStorage } from '../../../storage/local';
+
+export const ImportWalletCreationHeight = () => {
```
grod220, line 17: todo: See `needsOnboard()`. I wonder if we need a guard there. Maybe not since this is optional.

apps/extension/src/routes/page/onboarding/height.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  const handleClick = (event: MouseEvent | FormEvent) => {
+    handleSubmit(event).catch((error: unknown) => {
+      if (error instanceof Error) {
+        console.error('Error during submission:', error.message);
+      } else {
+        console.error('Unexpected error:', error);
+      }
+    });
+  };
```
grod220, line 40: Think this logic in the component can live in zustand too. Maybe all this creation height stuff should be its own zustand slice

apps/extension/src/wallet-services.ts
```diff
@@ -16,12 +16,16 @@ export const startWalletServices = async () => {
...
+  // Retrieve the wallet creation height flag from storage
+  const walletCreationBlockHeight = await localExtStorage.get('walletCreationBlockHeight');
+
   const services = new Services({
     grpcEndpoint,
     chainId: await getChainId(grpcEndpoint),
     walletId: WalletId.fromJsonString(wallet.id),
     fullViewingKey: FullViewingKey.fromJsonString(wallet.fullViewingKey),
     numeraires: numeraires.map(n => AssetId.fromJsonString(n)),
+    walletCreationBlockHeight: walletCreationBlockHeight!,
```
grod220, line 28: issue: let's not use `!` and handle the undefined case

grod220 (commented):
Great start to this!

JasonMHasperhoven:
suggestion: Did a quick draft, the idea is that copy to clipboard copies all (phrase + "\n blockheight: number") as most would probably do this
Frame 3

apps/extension/src/routes/page/onboarding/generate.tsx
TalDerei: as you pointed out, while this approach is non-blocking at the thread level,  it could still delay onboarding if the rpc calls take too long. Consequently, I think we should implement a timeout and a retry limit.

apps/extension/src/hooks/full-sync-height.ts
TalDerei: good call, I refactored `useSyncProgress` to call `fetchBlockHeight` to reduce code duplication.

apps/extension/src/routes/page/onboarding/generate.tsx
TalDerei: added code comment

apps/extension/src/routes/page/onboarding/generate.tsx
TalDerei: this was a great opportunity for me to refresh my knowledge about Zustand. I moved the suggested logic to a dedicated zustand store called `walletCreationBlockHeightSlice`, and `fetchBlockHeightWithFallback` and `fetchBlockHeight` are now utility functions.

apps/extension/src/routes/page/onboarding/generate.tsx
TalDerei: added type gaurd

apps/extension/src/routes/page/onboarding/generate.tsx
TalDerei: reworded comment

apps/extension/src/routes/page/onboarding/height.tsx
TalDerei: reworded comment

apps/extension/src/routes/page/onboarding/generate.tsx
TalDerei: references to the eslint rules implemented in https://github.com/penumbra-zone/web/pull/1721 that still need to be propagated to this repository.

apps/extension/src/routes/page/onboarding/generate.tsx
TalDerei: I moved the logic to the [zustand store](https://github.com/prax-wallet/web/pull/164/commits/4ecea54ab1deb4893e43da25b3cf37a18b902888#diff-47254f5f6d488f18262f49e33c899348386b370c99f0f9a7f1219160bf492196R20-R22) and propagated the error to the caller ([useEffect](https://github.com/prax-wallet/web/pull/164/commits/6ae261e2ea68362d90e5181fbe0c5b9f322ddf94#diff-0c7c024c9b3806be547cbd5b3d147f6d4c33e6a1e68814dda964ac666c8d606aR34) hook) which sets the error in the component's state. 

Screenshot 2024-09-01 at 10 36 39 AM

apps/extension/src/routes/page/onboarding/generate.tsx
TalDerei: modified logic to 

```ts
  const walletCreationBlockHeight = await fetchBlockHeight(randomGrpcEndpoint);

  if (walletCreationBlockHeight !== undefined) {
    return walletCreationBlockHeight;
  } else {
    // Remove the current endpoint from the list and retry with remaining endpoints
    const remainingEndpoints = endpoints.filter(endpoint => endpoint !== randomGrpcEndpoint);
    return fetchBlockHeightWithFallback(remainingEndpoints);
  }
```

It's essentially saying that if the tendermint client returns `undefined` or an error occurs, we want it will recursively attempt another endpoint, otherwise return the block height.

apps/extension/src/routes/page/onboarding/generate.tsx
TalDerei: great suggestion, looks much cleaner

Screenshot 2024-09-01 at 11 20 12 AM

apps/extension/src/routes/page/onboarding/height.tsx
TalDerei: "Enter the block height of your first Penumbra transaction" is inherently more involved than just simply using the term 'birthday height,' which is a more intuitive concept.

Maybe we should say something like `"Enter the Block Height at the time your wallet was created. This is your wallet's birthday height (Optional)."`?

apps/extension/src/routes/page/onboarding/height.tsx
TalDerei: opted to implement this, which required explicitly differentiating between fresh and existing wallets, as they require _different_ zustand slice interfaces.

apps/extension/src/routes/page/onboarding/height.tsx
TalDerei: modified to use `needsOnboard()`

apps/extension/src/routes/page/onboarding/height.tsx
TalDerei: created dedicated `zustand` slice and removed `handleClick` logic.

apps/extension/src/wallet-services.ts
TalDerei: modified type signature to `number | undefined` since existing wallets will have an undefined value for this field.

apps/extension/src/routes/page/onboarding/height.tsx
TalDerei: Screenshot 2024-09-01 at 8 41 08 PM

apps/extension/src/routes/page/onboarding/generate.tsx
TalDerei: _update_

Screenshot 2024-09-01 at 8 45 59 PM

apps/extension/src/routes/page/onboarding/height.tsx
TalDerei: _update_

Screenshot 2024-09-01 at 8 46 11 PM

apps/extension/src/routes/page/onboarding/generate.tsx
```diff
@@ -14,20 +14,38 @@ import { generateSelector } from '../../../state/seed-phrase/generate';
...
+        startCountdown();
+
+        if (!isInitialized.current && blockHeight === 0) {
+          await setBlockHeight(0, true);
+          isInitialized.current = true;
+        }
+      } catch (error) {
+        setError('Failed to fetch block height. Please try again later');
+      }
+    })();
```
TalDerei, line 47: target for reactQuery

grod220 (approved):
Post 🍐 approval 👍

apps/extension/src/routes/page/onboarding/generate.tsx
TalDerei: cc @turbocrime

apps/extension/src/hooks/full-sync-height.ts
TalDerei: modified to use react query pattern

apps/extension/src/routes/page/onboarding/generate.tsx
TalDerei: modified to use react query pattern

apps/extension/src/routes/page/onboarding/generate.tsx
TalDerei: confirmed proper error propagation with react query

Screenshot 2024-09-03 at 10 50 15 PM

apps/extension/src/routes/page/onboarding/generate.tsx
TalDerei: modified to incorporate more explicit error handling, rather than clobbering errors as undefined and recursing.

apps/extension/src/routes/page/onboarding/height.tsx
TalDerei: removed zustand slices

TalDerei:
tested various workflows and error conditions and looks good 👍

PR #165, opened 2024-08-19
Bundle mainnet genesis state

grod220:
Pairs with https://github.com/penumbra-zone/web/pull/1706


packages/context/src/index.ts
```diff
@@ -7,6 +7,9 @@ import { FullViewingKey, WalletId } from '@penumbra-zone/protobuf/penumbra/core/
 import { ChainRegistryClient } from '@penumbra-labs/registry';
 import { AppParameters } from '@penumbra-zone/protobuf/penumbra/core/app/v1/app_pb';
 import { AssetId } from '@penumbra-zone/protobuf/penumbra/core/asset/v1/asset_pb';
+import Penumbra1Genesis from './penumbra-1-genesis.json';
+import { CompactBlock } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/compact_block/v1/compact_block_pb';
+import { JsonValue } from '@bufbuild/protobuf';
```
TalDerei, line 12: Is there generally a distinction for using `JsonValue` over `JsonObject` to represent the genesis blob?

packages/context/src/genesis.test.ts
```diff
@@ -0,0 +1,10 @@
+import { describe, test, expect } from 'vitest';
+import Penumbra1Genesis from './penumbra-1-genesis.json';
```
TalDerei, line 2: organizationally, shouldn't extension static live in the `public` folder?

additionally, it seems that the json file is missing some fields: it's **483,361** lines long, compared to the **483,366** lines in this [genesis file](https://github.com/prax-wallet/web/pull/163/files)?

```
- "height": "0" (top)
-  "epochIndex": "0" (bottom)
- "altGasPrices": [] (bottom)
```

TalDerei (approved):

TalDerei:
closing https://github.com/prax-wallet/web/pull/163 in favor of this.

packages/context/src/genesis.test.ts
grod220: > organizationally, shouldn't extension static assets live in the public folder?

I don't have a strong opinion, but the context package seemed to be the easiest place as it sets up the services for the extension.

> it seems that the json file is missing some fields?

Oh strange 🤔 . I did a sync and logged the .toJson() of the genesis block. It seems like it excluded fields that had the default value (?). The result is the same, but not really sure why it did that. Can go with yours.

packages/context/src/index.ts
grod220: I suppose it doesn't matter given a `JsonObject` is a type of `JsonValue`. That said, it's more accurate to make this a `JsonObject` I'd say.

grod220:
@TalDerei feel free to take this over as it pairs with your other work

hdevalence:
Looking at the rust code here https://github.com/penumbra-zone/web/blob/main/packages/wasm/crate/src/view_server.rs#L116 , isn't the compact block passed into wasm as a binary-proto-encoded byte array? in that case, wouldn't it be best to store the compact block in that format to avoid serialization overhead, and which would also be smaller to include in the extension manifest?

TalDerei:
> @TalDerei feel free to take this over as it pairs with your other work

thanks taking over PR 👍

TalDerei:
> Looking at the rust code here https://github.com/penumbra-zone/web/blob/main/packages/wasm/crate/src/view_server.rs#L116 , isn't the compact block passed into wasm as a binary-proto-encoded byte array? in that case, wouldn't it be best to store the compact block in that format to avoid serialization overhead, and which would also be smaller to include in the extension manifest?

In theory, storing the compact block object in a binary-proto-encoded byte array format should be more efficient in terms of both serialization and space cost.

In terms of serialization, we'll still need to perform a deserialization operation, CompactBlock.fromBinary(new Uint8Array(binaryData)), since the block processor [processes](https://github.com/penumbra-zone/web/blob/main/packages/query/src/block-processor.ts#L172-L191) and perform checks on the in-memory compact block object before converting it to the binary-proto-encoded byte array that wasm consumes. However, deserializing from binary is more efficient than deserializing from JSON.

In terms of space, the compact block object for the genesis block is **~10 MB**, while its JSON representation is **~18 MB**. Surprisingly, when serialized to a binary-proto-encoded byte array using `compactBlock.toBinary()` and saved to a binary file, it takes up **~150 MB**.

Screenshot 2024-08-19 at 6 23 12 PM

packages/context/src/index.ts
TalDerei: agreed, changed to `JsonObject`

hdevalence:
I'm not sure where those numbers are coming from. The genesis compact block is 8035648 bytes. See https://github.com/penumbra-zone/penumbra/pull/4824 for code to generate the data

TalDerei:
> I'm not sure where those numbers are coming from. The genesis compact block is 8035648 bytes. See [penumbra-zone/penumbra#4824](https://github.com/penumbra-zone/penumbra/pull/4824) for code to generate the data

strange blowup, I was simply logging the binary representation of the compact block and manually saving to a file.

grod220:
Tested this out locally and things are working for me. @TalDerei after publishing the latest packages in penumbra-zone and pulling them in here, think you can merge.

In general, it's nice this saves the 8mb download. However, for a full sync the user will still have to download another +60mb. Connectrpc downloads all of those in parallel to the block sync logic. That said, it's not really noticed on my machine as the CPU/IO tasks are the bottleneck and those downloaded blocks are just waiting to be processed.

TalDerei:
> That said, it's not really noticed on my machine as the CPU/IO tasks are the bottleneck and those downloaded blocks are just waiting to be processed.

echoing the same sentiment on M3 pro

packages/context/src/index.ts
```diff
@@ -103,7 +104,15 @@ export class Services implements ServicesInterface {
       idbConstants: indexedDb.constants(),
     });
 
+    // Dynamically fetch binary file for genesis compact block
```
TalDerei, line 107: fetch api goes through the browser's network stack for local files, and there seems to be overhead in setting up the request, but latency is minimal

turbocrime (approved):

PR #172, opened 2024-08-20
Local storage v3 migration

grod220:
Closes https://github.com/prax-wallet/web/issues/166 & quite a bit more. We use chrome.local.storage for persisting some state for the extension. This includes things like wallets, grpc urls, frontend selection, etc. Given local storage is simply a key-value store, it does not come with any tooling to handle migrations. This has made this task a bit more complex than expected.

Our custom migration code for local storage was not quite robust enough to reliably handle the task of the issue. For that reason, I refactored some mechanics. Will add more color inline below. 

Note: This is PR is definitely within the "risky" territory. Migrating user state on the fly can have unintended consequences. This should be throughly tested before we ship another Prax version. For example, how does changing the user's rpc they saved in local storage affect an existing block sync in progress? It's very likely it won't take effect until the service worker is restarted (even though it shows otherwise in the extension). Hence the reason why changing RPCs triggers an extension restart.


apps/extension/src/storage/base.ts
```diff
@@ -20,30 +20,53 @@ export interface StorageItem<T> {
 }
 
 type Version = string;
+
+export type MigrationMap<OldState, NewState> = {
+  [K in keyof OldState & keyof NewState]?: (
+    prev: OldState[K],
+    get: <K extends keyof NewState>(key: K) => Promise<NewState[K]>,
```
grod220, line 27: Some migrations (in our case for V3), requires access to querying the store for other fields. Hence, the migration function is now given a `get` querier.

apps/extension/src/storage/base.ts
```diff
@@ -20,30 +20,53 @@ export interface StorageItem<T> {
...
+  ) => NewState[K] | Promise<NewState[K]>;
+};
+
 // It is quite difficult writing a generic that covers all migration function kinds.
 // Therefore, the writer of the migration should ensure it is typesafe when they define it.
-// See `migration.test.ts` for an example.
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-type Migration = Record<Version, (a: any) => any>;
-type Migrations<K extends string | number | symbol> = Partial<Record<K, Migration>>;
+export type Migrations<T> = Partial<Record<Version, MigrationMap<any, T>>>;
```
grod220, line 34: By far the biggest change. Before, migrations had to be:
- From the version of change to the most recent version (v1 -> v3). This means every time we added a new migration, we would have to go to those old versions and make sure they can properly be updated from say (v1 -> v4).  There was not concept of migrating one step at a time.
- The previous map had all migrations for all versions in one object. Was harder to understand (given we tend to think about migrations on a version 43 to version 44 kind of basis).

apps/extension/src/storage/base.ts
```diff
@@ -20,30 +20,53 @@ export interface StorageItem<T> {
...
 
     if (isEmptyObj(result)) {
       return this.defaults[key];
     } else {
-      return await this.migrateIfNeeded(key, result[key]);
+      if (skipMigration) {
+        return result[key].value;
+      } else {
+        return await this.migrateIfNeeded(key, result[key]);
+      }
```
grod220, line 69: Given some migrations need access to the wider storage getter, we have two ways to get from storage now.

apps/extension/src/storage/migrations/local-v1-migrations.ts
grod220, line 1: Copied over from old code.

apps/extension/src/storage/migrations/local-v2-migrations.ts
grod220, line 1: The core migration code that addresses the issue that prompted this whole task.

TalDerei:
Isn't it still the case that the recommended rpc entries in the prax registry contain outdated embedded frontends? I think we should remove any frontends that aren't using cron.

apps/extension/src/storage/base.ts
```diff
@@ -69,18 +94,47 @@ export class ExtensionStorage<T> {
...
+
+    // Check for an ongoing migration, if exists, return the result of that
+    const ongoingMigration = this.migrationLocks[String(key)] as Promise<T[K]> | undefined;
+    if (ongoingMigration) {
+      return ongoingMigration;
+    }
+
+    // Store the promise in the lock map. Ensures multiple callers don't run migrations multiple times.
+    const migrationPromise = this.migrateField(key, item);
+    this.migrationLocks[String(key)] = migrationPromise;
```
grod220, line 110: New addition. Given multiple parts of the codebase are accessing fields at the same time that require migration, we need a way to ensure multiple migrations are not run and thus share the same promise to be resolved.

grod220:
User flows to test:

- Frontend url is in the registry. After ext update: no change made.
- Frontend url is not in the registry. After ext update: changed to a random one in the registry.
- Grpc endpoint is in the registry. After ext update: no change made.
- Grpc endpoint is not in the registry. After ext update: changed to a random one in the registry.
  - Note, block syncing will still pull from previous grpc endpoint until service worker restarts. The UI will still display the old one.
- Go back to April extension version (before v2 migration). V1 -> v2 migration works for both frontend urls and grpc endpoints.

For all of these, confirm no corruption to the other keys (all get bumped to v3 after accessing (when migration takes place)).

TalDerei:
The migration is designed for version bumps, but I'm wondering if there's a way to rollback the extension (`V3 --> V2`), which seems to corrupt the local state storage?

Screenshot 2024-08-21 at 9 58 31 AM

apps/extension/src/storage/base.ts
```diff
@@ -20,30 +20,55 @@ export interface StorageItem<T> {
...
 // Therefore, the writer of the migration should ensure it is typesafe when they define it.
-// See `migration.test.ts` for an example.
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-type Migration = Record<Version, (a: any) => any>;
-type Migrations<K extends string | number | symbol> = Partial<Record<K, Migration>>;
+export type Migrations<T> = Partial<Record<Version, MigrationMap<any, T>>>;
+
+export type VersionSteps = Record<Version, Version>;
 
 export class ExtensionStorage<T> {
```
TalDerei, line 38: this is not a suggested code change, but rather a general statement if we decide to restructure our migration mechanism in the future. 

personally, migrations feel more transactional in nature, where migrations for multiple fields should ideally be handled atomically, rather than the current approach of performing migrations on individual fields separately.

one potential approach is a _batch_ migration mechanism, where you define migrations at the version level. It ensures atomicity by checking that all changes associated with a particular version are applied together in a single transaction. so either all migrations succeed or none do

apps/extension/src/storage/migrations/local-v2-migrations.ts
```diff
@@ -0,0 +1,78 @@
+import { LocalStorageState } from '../types';
+import { MigrationMap } from '../base';
+import { AppParameters } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/app/v1/app_pb';
+import { ChainRegistryClient } from '@penumbra-labs/registry';
+import { sample } from 'lodash';
+
+export const localV2Migrations: MigrationMap<LocalStorageState, LocalStorageState> = {
```
TalDerei, line 7: we support both sync / async migration functions – but it's not entirely clear to me how the current implementation explicitly handle errors that might occur during a migration. For instance, if the migration function fails, could this leave the local storage in a partially migrated or inconsistent state? Is there a retry / rollback mechanism?

apps/extension/src/storage/base.ts
```diff
@@ -20,30 +20,55 @@ export interface StorageItem<T> {
...
+  private migrationLocks: Record<string, Promise<T[keyof T]>> = {};
```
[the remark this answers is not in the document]
TalDerei: thinking more about migration locks, isn't there a risk of blocking if a migration gets stuck or slow? perhaps we should add a timeout mechanism to the migration locks

TalDerei:
tested some workflows – looks good to me

- Grpc endpoint and Frontend url is in the registry. After ext update: no change made.

Screenshot 2024-08-21 at 10 36 02 AM

Screenshot 2024-08-21 at 10 40 39 AM

----------------------------------------------------------------------------------------------------------------------------

- Frontend url is not in the registry. After ext update: changed to a random one in the registry.

Screenshot 2024-08-21 at 12 19 32 PM

Screenshot 2024-08-21 at 12 21 40 PM

----------------------------------------------------------------------------------------------------------------------------

- Grpc endpoint is not in the registry. After ext update: changed to a random one in the registry

Screenshot 2024-08-21 at 1 21 34 PM

Screenshot 2024-08-21 at 1 25 32 PM

----------------------------------------------------------------------------------------------------------------------------

- If connected a non-mainnet chain id, after ext update: no changes made to grpc endpoint.

Screenshot 2024-08-21 at 1 34 16 PM

Screenshot 2024-08-21 at 1 36 33 PM

apps/extension/src/storage/migrations/local-v2-migration.test.ts
```diff
@@ -0,0 +1,143 @@
...
+      await v2ExtStorage.set('frontendUrl', 'http://badfrontend.void');
+      const url = await v3ExtStorage.get('frontendUrl');
+      expect(url).not.toEqual('http://badfrontend.void');
+      expect(frontends.map(f => f.url).includes(url!)).toBeTruthy();
+    });
+  });
+
+  describe('grpcEndpoint', () => {
+    test('not set gets ignored', async () => {
+      await v1ExtStorage.set('grpcEndpoint', undefined);
```
TalDerei, line 95: what happens in scenarios where the `grpcEndpoint` or `frontendUrl` are _null_ rather than undefined?

TalDerei:
can you speak on partial storage migrations? suppose we're migrating from `V1` to `V3` through the storage versions, but the process partially fails after completing the migration to `V2`. what generally happens in this case?

apps/extension/src/storage/base.ts
grod220: This is an important comment. It's unfortunate that chrome.local.storage does not come with any support for versioning or migrations. For this reason, we had to create our own custom migration code. It was probably a bad design that each field has their own version key. A possibly better design is to save a global version number, check for needed migrations when someone attempts a `get()`, and if so, migrate all keys at once.

Going to spend today trying this out. The hardest part of this will be migrating from the old data structure to the new.

grod220:
Will pick up the missing code documentation tomorrow. Just pushed the update+tests that adopts the better global data structure for migrations in local storage. All is working on my end, but want to keep testing user flows given how sensitive of a change this is.

apps/extension/src/storage/base.ts
```diff
@@ -20,30 +20,55 @@ export interface StorageItem<T> {
 }
 
 type Version = string;
+
+export type MigrationMap<OldState, NewState> = {
+  [K in keyof OldState & keyof NewState]?: (
+    prev: OldState[K],
+    get: <K extends keyof NewState>(key: K) => Promise<NewState[K]>,
```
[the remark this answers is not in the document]
TalDerei: outdated type in favor of the simplified [MigrationFn](https://github.com/prax-wallet/web/pull/172/commits/8c8922c505bbe919b68e19fdbad026332d442f4f#diff-2bad0a69063227a5d56c8bb08e71d7de75d80fc3bd2e2106ed872311b1bfebc8R18)

apps/extension/src/storage/base.ts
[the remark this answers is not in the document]
TalDerei: **update:** the comment should be updated to reflect that the synchronization mechanism prevents concurrent modifications to the _entire_ storage object, not just a single field.

In terms of the risk of blocking, there now might be a risk of blocking between _storage versions_ if the migration gets stuck or slow for a specific storage object (rather than a partial storage key).

apps/extension/src/storage/base.ts
```diff
@@ -20,30 +20,55 @@ export interface StorageItem<T> {
...
   constructor(
     private storage: IStorage,
     private defaults: T,
     private version: Version,
-    private migrations: Migrations<keyof T>,
+    private migrations: Migrations<T> = {},
+    private versionSteps: VersionSteps = {},
   ) {}
 
   async get<K extends keyof T>(key: K): Promise<T[K]> {
```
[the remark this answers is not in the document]
TalDerei: **update:** `getter` and `setter` kick off the storage migration process for an entire _storage object_ at the version level.

- define a `Version` object with the target version for the storage and map of migrations `RequiredMigrations`, each entry corresponding to separate migration function,
- instantiate the `ExtensionStorage` class,
- call `get` or `set` methods to retrieve or set the values for a collection of keys,
- `get` method internally calls `withDbLock`, checking if the lock is already acquired,
- if the lock is not acquired (ie. `undefined`),  acquire the lock by assigning `dbLock` to the promise returned by `migrateOrInitializeIfNeeded`,
- wait for the lock to resolve, ensuring initialization or migration is complete,
  -  triggers `migrateOrInitializeIfNeeded` and optionally `migrateAllFields` to handle the migration,
- execute the internal get or set operation,
- finally, release the lock.

apps/extension/src/storage/migrations/local-v2-migrations.ts
TalDerei: **update:** : previously, partial state migrations allowed individual fields to have either synchronous or asynchronous migration functions, meaning a migration from one version to another could involve a mix of both sync and async functions. Now that we are migrating the entire storage object for a specified version, it raises a question: do all the fields being migrated have to be either sync or async?

this means that if a single field's migration function is async, the entire migration process needs to be treated as async. This currently doesn't seem to be explicitly enforced in the migration code. 

Additionally, the same error handling comment remains.

apps/extension/src/storage/base.ts
```diff
@@ -14,75 +15,71 @@ export interface IStorage {
...
-    prev: OldState[K],
-    get: <K extends keyof NewState>(key: K) => Promise<NewState[K]>,
-  ) => NewState[K] | Promise<NewState[K]>;
-};
+export interface RequiredMigrations {
+  // Explicitly require key '0' representing...
+  // It is quite difficult writing a generic that covers all migration function kinds.
+  // Therefore, the writer of the migration should ensure it is typesafe when they define it.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  0: MigrationFn<any, any>;
```
TalDerei, line 29: should we make this an _optional_ field? for instance, migration function for version 0 won't be relevant for a higher version.

apps/extension/src/storage/base.ts
```diff
@@ -14,75 +15,71 @@ export interface IStorage {
...
-export type VersionSteps = Record<Version, Version>;
+export interface ExtensionStorageProps<T extends { dbVersion: number }> {
+  storage: IStorage;
+  defaults: ExtensionStorageDefaults<T>;
+  version: Version;
+}
 
-export class ExtensionStorage<T> {
-  private migrationLocks: Record<string, Promise<T[keyof T]>> = {};
+export class ExtensionStorage<T extends { dbVersion: number }> {
```
TalDerei, line 45: can we include a description of the properties for the ExtensionStorage class:
- responsible for managing the storage system – storage versioning and migrations
- the state managed by this class explicitly includes versioning information from `dbVersion`

apps/extension/src/storage/migrations/local-v1-migration.ts
```diff
@@ -0,0 +1,132 @@
...
+  frontendUrl?: StorageItem<string | undefined>;
+  passwordKeyPrint?: StorageItem<KeyPrintJson | undefined>;
+  fullSyncHeight?: StorageItem<number | undefined>;
+  knownSites?: StorageItem<OriginRecord[]>;
+  params?: StorageItem<Stringified<AppParameters> | undefined>;
+  numeraires?: StorageItem<Stringified<AssetId>[]>;
+}
+
+// Update LocalStorageState to V1LocalStorageState if there is a version bump
+export const localV0Migration: MigrationFn<V0LocalStorageState, LocalStorageState> = v0 => {
```
TalDerei, line 37: `localV0Migration` handles the migration from the old `V0LocalStorageState` structure, where each field is wrapped in a `StorageItem`, to the new `LocalStorageState` structure. However, we don't have `localV1Migration: MigrationFn<LocalStorageState, LocalStorageState>` to test migrations within the `LocalStorageState` structure itself.

apps/extension/src/storage/base.ts
TalDerei: resolving as outdated

apps/extension/src/storage/base.ts
TalDerei: resolving as outdated

apps/extension/src/storage/base.ts
TalDerei: resolving as outdated

apps/extension/src/storage/base.ts
```diff
@@ -69,18 +92,29 @@ export class ExtensionStorage<T> {
   }
 
   private async migrateIfNeeded<K extends keyof T>(key: K, item: StorageItem<T[K]>): Promise<T[K]> {
```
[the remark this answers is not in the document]
TalDerei: resolving as outdated

apps/extension/src/storage/base.ts
```diff
@@ -69,18 +92,29 @@ export class ExtensionStorage<T> {
...
-    if (item.version !== this.version) {
-      const migrationFn = this.migrations[key]?.[item.version];
-      if (migrationFn) {
-        // Update the value to latest schema
-        const transformedVal = migrationFn(item.value) as T[K];
-        await this.set(key, transformedVal);
-        return transformedVal;
-      } else {
-        // If there's no migration function, handle it appropriately, possibly by just returning the current value
-        return item.value;
```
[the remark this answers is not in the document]
TalDerei: resolving as outdated

apps/extension/src/storage/base.ts
TalDerei: locking mechanism implementation changed, but rationale remains

TalDerei:
> tested some workflows – looks good to me

I still haven't retested these workflows post refactor

grod220:
Tested on all user flows, things in my view are looking good ✅

apps/extension/src/storage/base.ts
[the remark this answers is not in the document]
grod220: Talked offline. Not relevant to our specific migration, but think timeouts in the future could be a good feature to add.

apps/extension/src/storage/migrations/local-v2-migrations.ts
grod220: Talked offline. New storage datastructure does not mix async/sync migrations the way that the former field migrations did. All good here.

apps/extension/src/storage/migrations/local-v2-migration.test.ts
grod220: We don't set those as null, but null fields are possible (though, still of the opinion null is a programming language error)

TalDerei (approved):
left a minor comment; double-checked the locking system logic and that we're propagating the errors up the caller stack properly, and everything looks good to me.

TalDerei:
tested user workflows – looks good to me ✔️

* Grpc endpoint and Frontend url is in the registry. After ext update: no change made.

Screenshot 2024-08-24 at 3 16 54 PM

Screenshot 2024-08-24 at 3 19 15 PM

----------------------------------------------------------------------------------------------------------------------------

* Frontend url is not in the registry. After ext update: changed to a random one in the registry.

Screenshot 2024-08-24 at 3 32 33 PM

Screenshot 2024-08-24 at 3 35 32 PM

----------------------------------------------------------------------------------------------------------------------------

* Grpc endpoint is not in the registry. After ext update: changed to a random one in the registry

Screenshot 2024-08-24 at 6 27 43 PM

Screenshot 2024-08-24 at 6 29 40 PM

----------------------------------------------------------------------------------------------------------------------------
 
* If connected a non-mainnet chain id, after ext update: no changes made to grpc endpoint.

Screenshot 2024-08-24 at 6 50 53 PM

Screenshot 2024-08-24 at 6 53 38 PM

PR #173, opened 2024-08-22
inject request method

turbocrime:
temporary fix to mitigate missing request method

### DIFF
--- a/apps/extension/src/content-scripts/injected-penumbra-global.ts
+++ b/apps/extension/src/content-scripts/injected-penumbra-global.ts
@@ -57,6 +57,7 @@ class PraxInjection {
   private stateEvents = new EventTarget();
 
   private injection: Readonly<PenumbraProvider> = Object.freeze({
+    request: () => this.postConnectRequest().then(() => Promise.resolve()),
     connect: () => Promise.resolve(this.port ?? this.postConnectRequest()),
     disconnect: () => this.postDisconnectRequest(),
     isConnected: () => Boolean(this.port && this.presentState === PenumbraState.Connected),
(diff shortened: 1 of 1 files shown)


apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -57,6 +57,7 @@ class PraxInjection {
   private stateEvents = new EventTarget();
 
   private injection: Readonly<PenumbraProvider> = Object.freeze({
+    request: () => this.postConnectRequest().then(() => Promise.resolve()),
```
grod220, line 60: issue (blocking): can you add code documentation on the fact that this is temporary and should be removed after a pd point release?

TalDerei (approved):

apps/extension/src/content-scripts/injected-penumbra-global.ts
turbocrime: commented

turbocrime:
i think this should be part of a fix release, rather than waiting on another solution.

apps/extension/src/content-scripts/injected-penumbra-global.ts
turbocrime: this was approved by another review, but you've left a blocking comment, so i will avoid merging this unless you re-review.

go ahead and merge this if you intend to include it in a release.

grod220 (approved):
Merging to include in next release

grod220:
Will have to roll this one back. Looks like it works on first page load, but subsequent ones, it cannot make a connection: 

Screenshot 2024-08-26 at 11 50 49 AM

PR #189, opened 2024-09-05
onboarding: request level timeouts

TalDerei:
references https://github.com/prax-wallet/web/pull/164#discussion_r1739624507 cc @grod220 

based on offline discussion with @turbocrime, the channel transport has default [60 second timeout](https://github.com/penumbra-zone/web/blob/990291f6daef61bdb219f0c98bfbe481f25cfd83/packages/transport-dom/src/create.ts#L80), and grpc web transport have [no timeout](https://github.com/prax-wallet/web/blob/main/apps/extension/src/transport-options.ts#L4). The channel transport timeout is insufficient for our needs, as we may need to cycle through multiple rpcs to request the block height during onboarding. Instead, we’ve implemented a more granular request-level timeout of [5 seconds](https://github.com/prax-wallet/web/compare/channel-transport?expand=1#diff-1953ce36add59a143bebbfa0a7021d102fa0268e95a6fa9d4ab579ec8d7723afR38) directly within the fetch request.

### DIFF
--- a/apps/extension/src/hooks/latest-block-height.ts
+++ b/apps/extension/src/hooks/latest-block-height.ts
@@ -37,30 +37,16 @@ export const fetchBlockHeightWithTimeout = async (
   grpcEndpoint: string,
   timeoutMs = 5000,
 ): Promise<number> => {
-  return new Promise((resolve, reject) => {
-    const timeout = setTimeout(() => {
-      reject(new Error('Request timed out'));
-    }, timeoutMs);
+  const tendermintClient = createPromiseClient(
+    TendermintProxyService,
+    createGrpcWebTransport({ baseUrl: grpcEndpoint }),
+  );
 
-    const tendermintClient = createPromiseClient(
-      TendermintProxyService,
-      createGrpcWebTransport({ baseUrl: grpcEndpoint }),
-    );
-
-    tendermintClient
-      .getStatus({})
-      .then(result => {
-        if (!result.syncInfo) {
-          reject(new Error('No syncInfo in getStatus result'));
-        }
-        clearTimeout(timeout);
-        resolve(Number(result.syncInfo?.latestBlockHeight));
-      })
-      .catch(() => {
-        clearTimeout(timeout);
-        reject(new Error('RPC request timed out while fetching block height'));
-      });
-  });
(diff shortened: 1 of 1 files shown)


turbocrime:
> based on offline discussion with @turbocrime, the channel transport has default [60 second timeout](https://github.com/penumbra-zone/web/blob/990291f6daef61bdb219f0c98bfbe481f25cfd83/packages/transport-dom/src/create.ts#L80), and grpc web transport (intra-document requests) have [no timeout](https://github.com/prax-wallet/web/blob/main/apps/extension/src/transport-options.ts#L4)

i must have been unclear - that `internalTransportOptions` with `0` timeout is applied only to the transport used by extension pages querying rpc services, and loopback clients within the handlercontext.

none of the `createGrpcWebTransport` calls within the repository presently apply a timeout configuration, and i'm uncertain of the default as that implementation from a third party. i looked earlier but could not find a default, so it may also be `0`, which means it would ultimately be rejected by a browser-level HTTP failure.

kind of irrelevant i guess. reviewing now

apps/extension/src/hooks/latest-block-height.ts
turbocrime, line 1: i think this could be simplified.

notably, each `PromiseClient` method accepts an options parameter, including both `timeoutMs` and an `AbortSignal`.

```ts
interface CallOptions {
  /**
   * Timeout in milliseconds.
   *
   * Set to <= 0 to disable the default timeout.
   */
  timeoutMs?: number;

  /**
   * An optional AbortSignal to cancel the call.
   * If cancelled, an error with Code.Canceled is raised.
   */
  signal?: AbortSignal;
  
  headers?: HeadersInit;
  onHeader?(headers: Headers): void;
  onTrailer?(trailers: Headers): void;
  contextValues?: ContextValues;
}
```

pre-onboarding, you won't have access to the normal fullnode queriers from services. so constructing transports is definitely the right idea.

will come back with some suggestions.

turbocrime (approved):
suggestions in comments and in https://github.com/prax-wallet/web/pull/191

todo: the new multi-rpc fetcher seems overcomplicated, so it would be good to simplify.

merge when you're satisfied.

PR #191, opened 2024-09-05
suggestions for multiple rpc fetch

turbocrime:
n/t

### DIFF
--- a/apps/extension/src/hooks/latest-block-height.ts
+++ b/apps/extension/src/hooks/latest-block-height.ts
@@ -1,5 +1,4 @@
 import { useQuery } from '@tanstack/react-query';
-import { sample } from 'lodash';
 import { createPromiseClient } from '@connectrpc/connect';
 import { createGrpcWebTransport } from '@connectrpc/connect-web';
 import { TendermintProxyService } from '@penumbra-zone/protobuf';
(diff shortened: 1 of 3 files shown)


apps/extension/src/hooks/latest-block-height.ts
```diff
@@ -1,106 +1,63 @@
...
-};
-
-// Fetch the block height from a specific RPC endpoint.
-export const fetchBlockHeight = async (grpcEndpoint: string): Promise<number> => {
-  const tendermintClient = createPromiseClient(
-    TendermintProxyService,
-    createGrpcWebTransport({ baseUrl: grpcEndpoint }),
-  );
+    const latestBlockHeight = await tendermintClient
+      .getStatus({}, { signal: AbortSignal.timeout(timeoutMs) })
```
turbocrime, line 17: you could also use `timeoutMs` directly

```suggestion
      .getStatus({}, { timeoutMs })
```

this is fully unfounded, but i wonder if using the `timeoutMs` option might actually wait for the request to connect before starting the timer. a `signal` will definitely abort the request without any conditions.

TalDerei:
thanks for the suggestions, opted for a more minimal [diff](https://github.com/prax-wallet/web/pull/191/files#diff-1953ce36add59a143bebbfa0a7021d102fa0268e95a6fa9d4ab579ec8d7723afR16-R17)

PR #203, opened 2024-09-24
Re-enable numeraires in onboarding

grod220:
Bug fix. Now that a numeraire is [available in the registry](https://github.com/prax-wallet/registry/pull/94), Prax should be recognizing it and during onboarding, giving the users an opportunity to use it. This will enable usdc prices in minifront.

However, due to a react useEffect loading bug, it is skipping this step. This PR fixes that.

Screenshot 2024-09-24 at 9 45 29 PM

### DIFF
--- a/apps/extension/src/hooks/numeraires-query.ts
+++ b/apps/extension/src/hooks/numeraires-query.ts
@@ -3,26 +3,25 @@ import { ChainRegistryClient } from '@penumbra-labs/registry';
 import { useMemo } from 'react';
 
 export const useNumeraires = (chainId?: string) => {
-  const { data, isLoading, error, isError } = useQuery({
+  const { data, isLoading, isError } = useQuery({
     queryKey: ['registry', chainId],
     queryFn: async () => {
       const registryClient = new ChainRegistryClient();
       return registryClient.remote.get(chainId!);
     },
+    retry: 1,
+    retryDelay: 0,
     staleTime: Infinity,
     enabled: Boolean(chainId),
   });
 
   const numeraires = useMemo(() => {
-    if (!chainId) {
-      if (isError) {
-        console.error(`Could not load numeraires for chainId: ${chainId}`);
-      }
-      return [];
+    if (isError) {
+      console.error(`Could not load numeraires for chainId: ${chainId}`);
     }
 
     return data?.numeraires.map(n => data.getMetadata(n)) ?? [];
   }, [data, chainId, isError]);
 
-  return { numeraires, isLoading, error };
+  return { numeraires, isLoading, isError };
 };
--- a/apps/extension/src/shared/components/numeraires-form.tsx
+++ b/apps/extension/src/shared/components/numeraires-form.tsx
@@ -1,5 +1,4 @@
 import { AllSlices, useStore } from '../../state';
-import { useChainIdQuery } from '../../hooks/chain-id';
 import { useEffect, useState } from 'react';
 import { ServicesMessage } from '../../message/services';
 import { SelectList } from '@repo/ui/components/ui/select';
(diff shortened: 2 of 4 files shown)


apps/extension/src/hooks/numeraires-query.ts
```diff
@@ -3,26 +3,25 @@ import { ChainRegistryClient } from '@penumbra-labs/registry';
...
 export const useNumeraires = (chainId?: string) => {
-  const { data, isLoading, error, isError } = useQuery({
+  const { data, isLoading, isError } = useQuery({
     queryKey: ['registry', chainId],
     queryFn: async () => {
       const registryClient = new ChainRegistryClient();
       return registryClient.remote.get(chainId!);
     },
+    retry: 1,
+    retryDelay: 0,
```
grod220, line 13: A single retry feels ok, without specifying these, react-query tends to retry indefinitely.

apps/extension/src/shared/components/numeraires-form.tsx
```diff
@@ -13,39 +12,36 @@ const useNumerairesSelector = (state: AllSlices) => {
...
-  onSuccess: () => void | Promise<void>;
+  onSuccess: () => void;
 }) => {
-  const { chainId } = useChainIdQuery();
-  const { selectedNumeraires, selectNumeraire, saveNumeraires, networkChainId } =
-    useStore(useNumerairesSelector);
-
-  // 'chainId' from 'useChainIdQuery' is not available during onboarding,
-  // this forces you to use two sources to guarantee 'chainId' for both settings and onboarding
-  const { numeraires } = useNumeraires(chainId ?? networkChainId);
```
grod220, line 33: Instead of doing both (serving the settings & onboarding simultaneously internally), I changed the component so that chainId is passed in

apps/extension/src/shared/components/numeraires-form.tsx
```diff
@@ -13,39 +12,36 @@ const useNumerairesSelector = (state: AllSlices) => {
...
 
+  // If query errors or there aren't numeraires in the registry, can skip
   useEffect(() => {
-    if (numeraires.length === 0) {
-      void onSuccess();
+    if (isError || (!isLoading && numeraires.length === 0)) {
+      onSuccess();
     }
-  }, [numeraires]);
+  }, [numeraires.length, isError, isLoading]);
```
grod220, line 35: The previous source of the bug. During the loading state, the numeraires array was length zero. Hence it was being skipped. Now it only skips if there is an error or it's still loading.

apps/extension/src/shared/components/numeraires-form.tsx
```diff
@@ -13,39 +12,36 @@ const useNumerairesSelector = (state: AllSlices) => {
...
 
   const [loading, setLoading] = useState(false);
 
   const handleSubmit = () => {
     setLoading(true);
     void (async function () {
       await saveNumeraires();
-      await chrome.runtime.sendMessage(ServicesMessage.ChangeNumeraires);
-      await onSuccess();
+      void chrome.runtime.sendMessage(ServicesMessage.ChangeNumeraires);
```
grod220, line 43: Does not feel necessary to await the response here, hold up the UI from progressing. The fire-and-forget works.

TalDerei (approved):
issue: the onboarding process can be completed _without_ selecting a token denomination. Should we enforce a selection?

Screenshot 2024-09-25 at 7 39 52 PM

apps/extension/src/hooks/numeraires-query.ts
grod220: Hmm. Maybe it's the delay then that causes such long retry times. Alas, for our case, we want to fail fast given this is not a critical setting to select.

grod220 (commented):
> Should we enforce a selection?

Choosing a numeraire does not feel like a critical setting selection. If there is a fetching error, we want to skip it and not fail hard at this step. Think we are ok to keep it as optional.

PR #208, opened 2024-10-07
fix: allow http for production prax too

VanishMax:
Allows `http` apps to connect to production-built Prax


grod220 (approved):
question: I'm surprised this didn't break any tests 🤔 . Can you check on the suite to ensure coverage is good?

vacekj (approved):

PR #209, opened 2024-10-11
fix: beta extension bundle

VanishMax:
Relates to https://github.com/penumbra-zone/web/issues/1811

The extension relies on the `.env` file that sets the `PRAX` variable with the chrome extension ID. This ID is then used to correctly inject the provider to web pages. The build process didn't consider it, and the BETA extension injected the prod ID. The `client` package, in its turn, always checks the injected id matches chrome manifest to avoid security breaches, – it all means the BETA package didn't work completely.

This PR fixes the issue with BETA extension not being built correctly:
- Separates webpack bundles into the `prod` and `beta` versions
- Removes the `.env` files in favor of string replacements
- Beta bundle produces the `beta-dist` directory
- Beta bundle also modifies the `manifest.json` file to update favicons and info about the extension

### DIFF
--- a/apps/extension/webpack.prod-config.ts
+++ b/apps/extension/webpack.prod-config.ts
@@ -1,0 +1,17 @@
+/**
+ * This config overrides the env file for the extension and changes the output directory
+ */
+
+import dotenv from 'dotenv';
+import path from 'node:path';
+import config from './webpack.config.js';
+
+dotenv.config({ path: '.env', override: true });
+
+const __dirname = new URL('.', import.meta.url).pathname;
+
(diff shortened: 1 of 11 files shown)


apps/extension/webpack.prod-config.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * This config overrides the env file for the extension and changes the output directory
+ */
+
+import dotenv from 'dotenv';
+import path from 'node:path';
+import config from './webpack.config.js';
+
+dotenv.config({ path: '.env', override: true });
```
grod220, line 9: question: would it make it simpler if we didn't have an env file at all and just put the ids in the respective prod vs beta webpack configs?

grod220 (approved):

Issue #206, opened 2024-10-03
Noble forwarding addrs


grod220:
Implement the web version in Prax of https://github.com/penumbra-zone/penumbra/pull/4878. Noble launching feature Early Nov.

Follows this userflow:

![image](https://github.com/user-attachments/assets/3113d936-063c-4014-8378-fd657e91994c)

## Getting sequence numbers

There are three RPC responses to the register call:
1. unregistered, no deposits (failure 1)
2. registered already (failure 2)
3. unregistered, deposit exists (success)

Should use binary search to find the next available sequence number for user. Prax can store it directly and the latest sequence number can be found on demand with at most 16 RPC requests

Let's say someone has already generated 50 addresses. Any sequence <=50 will result in an RPC success (aka already deposited). Anything >50 will return an error. So we keep splitting the address space in half, checking for success/error until we come down to two (50 success, 51 error).

## Proposed design

![screenshot_2024-09-24_at_11 50 08___am_720](https://github.com/user-attachments/assets/30cb7ce9-56b3-48f3-ac2f-ee1ff8d6d7ed)

States to consider:
- If toggling the noble one, should replace the identicon with Noble icon + Noble address
- Likely should ditch the incognito icons
- Have a disable UI state so two can't be toggled at once
- Requires refactoring this component
- Needs loading/loaded/error states

TalDerei:
This should be moved back into the backlog – according to henry, "forwarding addresses may have slipped schedule and might be pushed to next year."

PR #215, opened 2024-10-15
Noble sequence search

grod220:
Related to: https://github.com/prax-wallet/prax/issues/206 & https://github.com/penumbra-zone/web/pull/1855 & https://github.com/penumbra-zone/penumbra/pull/4878

Adds a new Noble client and sequence searcher. Next PR will focus on implementing UI to use within Prax account selector.


grod220:
Ready for review, but dependent on https://github.com/penumbra-zone/web/pull/1855

packages/noble/src/client.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  async registerAccount({ sequence, accountIndex }: { sequence: number; accountIndex?: number }) {
+    const { penumbraAddr, nobleAddrBech32, nobleAddrBytes } = getNobleForwardingAddr(
+      sequence,
+      this.fvk,
+      this.channel,
+      accountIndex,
+    );
+
+    console.log({ penumbraAddr: bech32mAddress(penumbraAddr), nobleAddrBech32 });
```
JasonMHasperhoven, line 53: Is this intended?

JasonMHasperhoven (approved):

JasonMHasperhoven (changes requested):
Checks are failing

TalDerei (commented):
I think we should wait to merge until there's clarity around the notion of sequence numbers wrapping and what that means in practice. Specifically, if the system exhausts all sequence numbers and begins reusing them, are we certain that user deposits won't be lost or rejected due to an old address being mistakenly reused?

packages/noble/src/sequence-search.ts
```diff
@@ -0,0 +1,78 @@
+import { FullViewingKey } from '@penumbra-zone/protobuf/penumbra/core/keys/v1/keys_pb';
+import { NobleClientInterface, NobleRegistrationResponse } from './client';
+
+// Search space (sequence number) is 2 bytes wide
+export const MAX_SEQUENCE_NUMBER = 65535;
+
+// Perform binary search to find the earliest unused noble sequence number
+export const getNextSequence = async ({
```
TalDerei, line 8: Is there a reason the binary search for obtaining a noble sequence number wasn't implemented as a wasm helper?

TalDerei:
> Next PR will focus on implementing UI to use within Prax account selector.

flagging for external reviewers: the UI component won't be included until the noble one-time deposit addresses land on Noble mainnet in approximately two weeks.

package.json
```diff
@@ -51,5 +51,57 @@
...
+        "@penumbra-zone/query",
+        "@penumbra-zone/services",
+        "@penumbra-zone/storage",
+        "@penumbra-zone/transport-chrome",
+        "@penumbra-zone/transport-dom",
+        "@penumbra-zone/types",
+        "@penumbra-zone/ui",
+        "@penumbra-zone/wasm"
+      ]
+    }
```
TalDerei, line 105: we shouldn't commit these overrides in the package.json

TalDerei (approved):

packages/noble/src/sequence-search.ts
grod220: Because we don't really have any client fetching infra in wasm for making grpc or http requests. We could investigate this, but I'm not sure if it would add value here. Do you disagree?

packages/noble/src/client.ts
TalDerei: additional `console.log(res);`

debug logs?

grod220: Will remove 👍

packages/noble/src/sequence-search.ts
TalDerei: iirc, I thought core exposes an external rust interface that handles the networking logic that wasm can consume? maybe I'm wrong, but I don't disagree that it doesn't necessarily add extra value here.

packages/noble/src/sequence-search.ts
```diff
@@ -0,0 +1,78 @@
...
+  right,
+  client,
+  fvk,
+  accountIndex,
+}: {
+  left: number;
+  right: number;
+  client: NobleClientInterface;
+  fvk: FullViewingKey;
+  accountIndex?: number;
```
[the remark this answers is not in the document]
grod220: The account index is only used on our side to generate the proper forwarding address. The function to generate the noble address would throw if it something were wrong with the account index.

packages/noble/src/client.ts
```diff
@@ -0,0 +1,106 @@
...
+          gasLimit: 200000n,
+        },
+      },
+      signatures: [new Uint8Array()],
+    });
+
+    const client = await StargateClient.connect(this.endpoint);
+
+    try {
+      const res = await client.broadcastTx(tx.toBinary());
```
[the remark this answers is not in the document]
grod220: The caller of this client method would be the ones responsible for re-tries I'd say. It's also easy enough for the user to re-request a noble address if an error is thrown. Can revisit this topic though as the UI comes together and we see how that would work.

packages/noble/src/sequence-search.ts
grod220: Given the use of tokio, there is not a wasm-compatible request helper I know of.

packages/noble/src/sequence-search.ts
```diff
@@ -0,0 +1,78 @@
...
+}): Promise<number | undefined> => {
+  // Entire search sequence space has been exhausted
+  if (left > right) {
+    return undefined;
+  }
+
+  const mid = Math.floor((left + right) / 2);
+  const response = await client.registerAccount({ sequence: mid, accountIndex });
```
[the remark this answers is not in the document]
grod220: - On no deposited funds, both requests get a `NeedsDeposit` response back
- On deposited funds, the first request gets a `Success` and the second a `AlreadyRegistered`
- On already registered, both get `AlreadyRegistered`

That said, don't see a case when multiple clients are in use at the same time

grod220:
@JasonMHasperhoven given your first approval, going to interpret good to merge given checks are passing now 👍

Issue #214, opened 2024-10-14
beta testing: handle wallet birthday height for non-mainnet chain ids


TalDerei:
from @grod220: "noticed with the beta testers they were not receiving funds when they go through the new seedphrase flow in onboarding for testnet. But then realized that walletCreationBlockHeight is before the rpc selection screen and derived from the rpcs in the registry (all mainnet). That means their wallet creation height was higher than the latest block height on testnet, and therefore were skipping block syncing. Think we should do two things:

- On rpc selection screen, if we see they choose an rpc with a non-mainnet chain id, we should delete or recalculate walletCreationBlockHeight
- On advanced -> clear cache, we should delete walletCreationBlockHeight in storage (or accept an option to delete creation height in local storage)"

grod220:
Along with this work: we should use the fetched registry and not the bundled one (only as a backup if fetch fails)

PR #216, opened 2024-10-16
handle wallet birthday height for non-mainnet chain ids

TalDerei:
references https://github.com/prax-wallet/prax/issues/214

unblocks beta testers to receive test funds


vacekj (approved):

apps/extension/src/routes/page/onboarding/success.tsx
grod220, line 1: issue: This actually too late. The wallet birthday at this point has already been passed to the block processor. It will only pick up this change after the service worker resets.

grod220 (changes requested):
I think I've realized that the core of the issue can be fixed be re-ordering the onboarding steps.

So instead of:
```
New seedphrase + birthday -> password -> rpc -> frontend -> numeraire
```
We can do:
```
rpc -> New seedphrase + birthday -> password -> frontend -> numeraire
```

That way, things downstream that require knowledge of the chain id don't have to assume mainnet. This prevents the need for us to have steps after to correct incorrect state.

_Edit: In this case, we couldn't forget that we have to hold services init with the import-seed-phrase route completes the birthday step._

apps/extension/src/routes/page/onboarding/success.tsx
TalDerei: this is a great catch. I was under the false impression that block syncing started after the last onboarding step, but it actually begins right after the user selects their rpc endpoint.

TalDerei:
> I think I've realized that the core of the issue can be fixed be re-ordering the onboarding steps.

This is tricky. I've considered this, but separating the rpc and frontend steps doesn't feels natural to the onboarding flow in my opinion.

apps/extension/src/routes/page/onboarding/set-grpc-endpoint.tsx
```diff
@@ -3,11 +3,26 @@ import { FadeTransition } from '@repo/ui/components/ui/fade-transition';
...
+  const onSuccess = async (): Promise<void> => {
+    await new Promise(resolve => setTimeout(resolve, 1000));
+
+    const storedParams = await localExtStorage.get('params');
+    if (storedParams) {
+      const parsedParams = JSON.parse(storedParams) as AppParameters;
+      if (!parsedParams.chainId.includes('penumbra-1')) {
+        await localExtStorage.set('walletCreationBlockHeight', 0);
+      }
+    }
```
TalDerei, line 23: retrieves the app parameters and checks the associated chainId. 

An additional pre-step should be added, referencing https://github.com/penumbra-zone/web/pull/1853, to check if the connected wallet is the beta wallet.

apps/extension/src/routes/page/onboarding/set-grpc-endpoint.tsx
```diff
@@ -3,11 +3,26 @@ import { FadeTransition } from '@repo/ui/components/ui/fade-transition';
...
 
-  const onSuccess = (): void => {
+  // For beta testing: set the wallet block height to zero for non-mainnet chain IDs.
+  // This logic only runs after the user selects their rpc endpoint.
+  const onSuccess = async (): Promise<void> => {
+    await new Promise(resolve => setTimeout(resolve, 1000));
+
+    const storedParams = await localExtStorage.get('params');
+    if (storedParams) {
+      const parsedParams = JSON.parse(storedParams) as AppParameters;
```
grod220, line 19: issue: we should parse using buf build to parse:

```ts
const parsedParams = AppParameters.fromJsonString(storedParams);
```

apps/extension/src/routes/page/onboarding/set-grpc-endpoint.tsx
```diff
@@ -3,11 +3,26 @@ import { FadeTransition } from '@repo/ui/components/ui/fade-transition';
...
+import { localExtStorage } from '../../../storage/local';
+import { AppParameters } from '@penumbra-zone/protobuf/penumbra/core/app/v1/app_pb';
 
 export const SetGrpcEndpoint = () => {
   const navigate = usePageNav();
 
-  const onSuccess = (): void => {
+  // For beta testing: set the wallet block height to zero for non-mainnet chain IDs.
+  // This logic only runs after the user selects their rpc endpoint.
+  const onSuccess = async (): Promise<void> => {
```
grod220, line 14: issue: unfortunately, I still think this is running after services has initialized. Sadly it looks like the `useGrpcEndpointForm` is incredibly complicated. It's an example of where useEffect() adds complexity... See:

- [setGrpcEndpoint](https://github.com/prax-wallet/prax/blob/3ee1ff6954677d0862623c930f3ee985069496a2/apps/extension/src/shared/components/grpc-endpoint-form/use-grpc-endpoint-form.ts#L132-L138) runs before onSuccess
- which means local storage [being set first](https://github.com/prax-wallet/prax/blob/3ee1ff6954677d0862623c930f3ee985069496a2/apps/extension/src/state/network.ts#L25)
- which means [startWalletServices()](https://github.com/prax-wallet/prax/blob/3ee1ff6954677d0862623c930f3ee985069496a2/apps/extension/src/wallet-services.ts#L16) proceeds through that line and starts services with the old walletCreationBlockHeight

TalDerei: very tricky – only thing that comes to mind as a potential solution is refactoring the `useEffect` to delay service initialization using a flag where `startWalletServices` only runs after `walletCreationBlockHeight` is set.

TalDerei: or maybe move the logic directly into [onSubmit](https://github.com/prax-wallet/prax/blob/3ee1ff6954677d0862623c930f3ee985069496a2/apps/extension/src/shared/components/grpc-endpoint-form/use-grpc-endpoint-form.ts#L112)?

TalDerei: like why can't we reorder the operations inside the [onSubmit](https://github.com/prax-wallet/prax/blob/3ee1ff6954677d0862623c930f3ee985069496a2/apps/extension/src/shared/components/grpc-endpoint-form/use-grpc-endpoint-form.ts#L112) async function?

```ts
        const storedParams = await localExtStorage.get('params');
        if (storedParams) {
          const parsedParams = JSON.parse(storedParams) as AppParameters;
          if (!parsedParams.chainId.includes('penumbra-1')) {
            await localExtStorage.set('walletCreationBlockHeight', 0);
          }
        }
    

       await onSuccess();
```
    
which ensures a proper sequencing?

TalDerei: there must be a way to explicitly block on wallet service init until this code block completes.

TalDerei: or maybe move `await onSuccess()` before `void chrome.runtime.sendMessage(ServicesMessage.ClearCache)`? does that functionally break anything?

TalDerei: or we can block wallet service init directly inside [onboardGrpcEndpoint](https://github.com/prax-wallet/prax/blob/3ee1ff6954677d0862623c930f3ee985069496a2/apps/extension/src/wallet-services.ts#L16)?

TalDerei: settled on https://github.com/prax-wallet/prax/pull/216/commits/fcd8e41e16de32ac4916e6a2ece2e48916543ea4 for simplicity and readability.

apps/extension/src/routes/page/onboarding/set-grpc-endpoint.tsx
TalDerei: outdated

apps/extension/src/routes/page/onboarding/success.tsx
```diff
@@ -2,10 +2,22 @@ import { Button } from '@repo/ui/components/ui/button';
...
 import { useStore } from '../../../state';
 import { getDefaultFrontend } from '../../../state/default-frontend';
+import { localExtStorage } from '../../../storage/local';
 
 export const OnboardingSuccess = () => {
   const defaultFrontendUrl = useStore(getDefaultFrontend);
 
+  // Conditional: for beta-testing purposes, set the wallet block height to zero for non-mainnet chain id's.
+  void (async () => {
+    const storedParams = await localExtStorage.get('params');
```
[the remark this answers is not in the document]
TalDerei: outdated

apps/extension/src/routes/page/onboarding/set-grpc-endpoint.tsx
TalDerei: refactored, outdated

TalDerei:
how can we test this with the beta chrome extension per [penumbra-zone/web#1853](https://github.com/penumbra-zone/web/pull/1853)? cc @VanishMax

grod220:
> how can we test this with the beta chrome extension

Whenever we want, we can publish to the prax staging env:

https://chromewebstore.google.com/detail/prax-wallet-beta/ejpfkiblcablembkdhcofhokccbbppnc?authuser=0&hl=en

We should probably create a github action that we can manually run to help us with this

grod220:
Alternative suggestion for using a `beforeSubmit` hook: https://github.com/prax-wallet/prax/pull/220

grod220 (approved):
Could you write a few unit tests for `correctBirthdayHeightIfNeeded()`

grod220 (approved):

PR #217, opened 2024-10-17
fix(extension): #207: fix resetting the fullSyncHeight in local storage after cache clearing

### LINKED ISSUE #207: Clearing cache & block sync state
When clearing cache, the last block synced status does not clear until a new block is synced. This can cause feelings of unresponsiveness. In case of mainnet, genesis state sync can take some time so it accentuate this feeling. Example:

https://github.com/user-attachments/assets/d8626292-c949-418c-ae92-26ce7071d61e

Likely need to explicitly reset the chrome.storage.local value for lastBlockSynced when issuing a cache clear action.

VanishMax:
Fixes #207 

I understand that the issue was bigger but at least this PR fixes the symptom of incorrect cache clearing, which is the sync bar getting stuck in the incorrect state.

Read Tal's comment https://github.com/prax-wallet/prax/issues/207#issuecomment-2413688942 on the "data objects could not be cloned" error – we should perhaps plan the specific set tasks to deal with it. 

However, for now I decided that it might not hurt to add a "reload page" button next to the error messages in Minifront: https://github.com/penumbra-zone/web/pull/1861

### DIFF
--- a/apps/extension/src/routes/popup/settings/settings-clear-cache.tsx
+++ b/apps/extension/src/routes/popup/settings/settings-clear-cache.tsx
@@ -11,15 +11,14 @@ import { SettingsScreen } from './settings-screen';
 const useCacheClear = () => {
   const navigate = usePopupNav();
   const [loading, setLoading] = useState(false);
+  const setFullSyncHeight = useStore(state => state.network.setFullSyncHeight);
 
   const handleCacheClear = () => {
     setLoading(true);
 
     void (async function () {
+      await setFullSyncHeight(0);
       await chrome.runtime.sendMessage(ServicesMessage.ClearCache);
-      useStore.setState(state => {
-        state.network.fullSyncHeight = undefined;
-      });
       navigate(PopupPath.INDEX);
     })();
   };
--- a/apps/extension/src/routes/popup/settings/settings-clear-cache.tsx
+++ b/apps/extension/src/routes/popup/settings/settings-clear-cache.tsx
@@ -17,7 +17,7 @@ const useCacheClear = () => {
     setLoading(true);
 
     void (async function () {
-      await setFullSyncHeight(undefined);
+      await setFullSyncHeight(0);
       await chrome.runtime.sendMessage(ServicesMessage.ClearCache);
       navigate(PopupPath.INDEX);
     })();
--- a/apps/extension/src/routes/popup/settings/settings-clear-cache.tsx
+++ b/apps/extension/src/routes/popup/settings/settings-clear-cache.tsx
@@ -11,14 +11,15 @@ import { SettingsScreen } from './settings-screen';
 const useCacheClear = () => {
   const navigate = usePopupNav();
   const [loading, setLoading] = useState(false);
-  const setFullSyncHeight = useStore(state => state.network.setFullSyncHeight);
 
   const handleCacheClear = () => {
     setLoading(true);
 
     void (async function () {
-      await setFullSyncHeight(0);
       await chrome.runtime.sendMessage(ServicesMessage.ClearCache);
+      useStore.setState(state => {
+        state.network.fullSyncHeight = undefined;
+      });
       navigate(PopupPath.INDEX);
     })();
   };
--- a/apps/extension/src/routes/popup/settings/settings-clear-cache.tsx
+++ b/apps/extension/src/routes/popup/settings/settings-clear-cache.tsx
@@ -17,7 +17,7 @@ const useCacheClear = () => {
     setLoading(true);
 
     void (async function () {
-      await setFullSyncHeight(0);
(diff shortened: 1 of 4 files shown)


apps/extension/src/routes/popup/settings/settings-clear-cache.tsx
```diff
@@ -11,15 +11,14 @@ import { SettingsScreen } from './settings-screen';
 const useCacheClear = () => {
   const navigate = usePopupNav();
   const [loading, setLoading] = useState(false);
+  const setFullSyncHeight = useStore(state => state.network.setFullSyncHeight);
 
   const handleCacheClear = () => {
     setLoading(true);
 
     void (async function () {
+      await setFullSyncHeight(0);
```
grod220, line 20: question: Does a height synced of `0` mean that the genesis state has been synced? I actually think the data structure is more like:

```
fullSyncHeight = undefined --> no syncing has been completed
fullSyncHeight = 0 --> genesis state synced
fullSyncHeight = <number> --> that block number has been synced
```

Can you validate this? May help checking the logic of the loading bar to see what text shows under which condition.

VanishMax: The are no checks like `if (typeof fullSyncHeight === 'undefined'), only `!fullSyncHeight`. This way both `0` and `undefined` mean loading, `number` means syncing, and `latestKnownBlockHeight - fullSyncHeight <= 10` means synced

VanishMax: Updated zeros to `undefined`

TalDerei (approved):
to reiterate, this sufficiently addresses the symptom, but doesn't fully resolve the issue as the root cause is deeper. Alternatively, we could modify the [promise](https://github.com/prax-wallet/prax/blob/main/apps/extension/src/wallet-services.ts#L89) in the message listener with `await Promise.allSettled([localExtStorage.remove('params'), indexedDb.clear(), chrome.storage.local.remove('fullSyncHeight')])`, but they're functionally identical. 

Specifically, the deeper problem lies in the fact that the [listeners](https://github.com/prax-wallet/prax/blob/main/apps/extension/src/wallet-services.ts#L33) responsible for initializing the content scripts are blocked by the service worker while it synchronizes the [genesis](https://github.com/prax-wallet/prax/blob/main/apps/extension/src/wallet-services.ts#L31) block. I believe there's no way to persist the wallet's connection state across page refreshes or extension reloads because the [MessagePort](https://github.com/prax-wallet/prax/blob/main/apps/extension/src/content-scripts/injected-penumbra-global.ts#L73) is an in-memory object. So when the service worker restarts via [chrome.runtime.reload()](https://github.com/prax-wallet/prax/blob/main/apps/extension/src/wallet-services.ts#L92), it's waiting for the `MessagePort` to be initialized, and redirects to the onboarding screen until that's finished.

apps/extension/src/routes/popup/settings/settings-clear-cache.tsx
```diff
@@ -11,15 +11,14 @@ import { SettingsScreen } from './settings-screen';
 const useCacheClear = () => {
   const navigate = usePopupNav();
   const [loading, setLoading] = useState(false);
+  const setFullSyncHeight = useStore(state => state.network.setFullSyncHeight);
 
   const handleCacheClear = () => {
     setLoading(true);
 
     void (async function () {
+      await setFullSyncHeight(undefined);
```
grod220, line 20: issue: I don't think this will work as doing `chrome.storage.set({ "fullSyncHeight": undefined })` is a no-op. Can you confirm with a test? May help reviewing the docs as well: https://developer.chrome.com/docs/extensions/reference/api/storage#type-StorageArea.

VanishMax: Truly has some problems, reverted to the Tal's initial solution with `allSettled`

apps/extension/src/wallet-services.ts
```diff
@@ -86,7 +86,11 @@ const attachServiceControlListener = ({
       case ServicesMessage.ClearCache:
         void (async () => {
           blockProcessor.stop('clearCache');
-          await Promise.allSettled([localExtStorage.remove('params'), indexedDb.clear()]);
+          await Promise.allSettled([
+            localExtStorage.remove('params'),
+            indexedDb.clear(),
+            chrome.storage.local.remove('fullSyncHeight'),
```
grod220, line 92: Ah, unfortunately using the chrome storage api directly is not typesafe and could cause bugs with old Prax versions that have not undergone db migrations. Can you follow up with a PR to do:

```diff
- chrome.storage.local.remove('fullSyncHeight'),
+ localExtStorage.remove('fullSyncHeight'),
```

PR #224, opened 2024-11-01
prep v14.1 release

TalDerei:
@grod220 question about the `change RPC & re-sync` user-flow: should changing the rpc always trigger genesis sync? Currently, it does not.


grod220:
Is it because changing the RPC will force it to still observe the walletCreationHeight? I wonder what the best way to handle that is. If it's the same chain, it's all good. But if not, that is a problem. I suppose given we'll soon remove the ability for folks to add a wallet birthday, we should clear it on rpc changes.

TalDerei:
> Is it because changing the RPC will force it to still observe the walletCreationHeight?

exactly,  can we incorporate this into the scope of [prax-wallet/prax#226](https://github.com/prax-wallet/prax/issues/226)?

grod220:
Added to description 👍

PR #241, opened 2024-11-22
views: improve approval dialogue rendering

TalDerei:
references https://github.com/prax-wallet/prax/issues/240

I opted _not_ to import the new `valueViewComponent` since the UI V2 library is still in-flight and not finalized. To avoid adding further complexity — especially given the existing lack of uniformity across UI libraries (deprecated V1, V2, and the prax-specific UI) — I reused the existing `valueViewComponent`.

According to the developer docs, the maximum size for an extension popup is limited to [800x600](https://developer.chrome.com/docs/extensions/mv2/reference/browserAction#:~:text=The%20popup%20cannot%20be%20smaller,file%20with%20the%20popup's%20contents) unfortunately.

Moreover, I took some design liberties in the approval dialogue by refactoring the css and tightening the dimensions, but this probably goes _beyond_ the scope of what was intended. 

-----------------------

_before_

https://github.com/user-attachments/assets/917d9ef9-4cd9-4a0e-8e6f-0c3001646401

-----------------------

_after_ 

https://github.com/user-attachments/assets/9e862be5-4bab-4a1b-a8c5-76048d2e8b3c

PR #242, opened 2024-11-25
prax: add non-native-fee-warning

TalDerei:
second part of https://github.com/penumbra-zone/web/issues/1882

focused more on functionality rather than design, which definitely has room for improvement.

---------------------------

Screenshot 2024-11-24 at 10 50 22 PM

### DIFF
--- a/apps/extension/src/routes/popup/approval/transaction/index.tsx
+++ b/apps/extension/src/routes/popup/approval/transaction/index.tsx
@@ -11,7 +11,7 @@ import type { Jsonified } from '@penumbra-zone/types/jsonified';
 import { TransactionViewTab } from './types';
 import { ChainRegistryClient } from '@penumbra-labs/registry';
 import { viewClient } from '../../../../clients';
-import { useState } from 'react';
+import { TransactionView } from '@penumbra-zone/protobuf/penumbra/core/transaction/v1/transaction_pb';
 
 const getMetadata: MetadataFetchFn = async ({ assetId }) => {
   const feeAssetId = assetId ? assetId : new ChainRegistryClient().bundled.globals().stakingAssetId;
@@ -20,13 +20,21 @@ const getMetadata: MetadataFetchFn = async ({ assetId }) => {
   return denomMetadata;
 };
 
+const hasAltGasFee = (txv?: TransactionView): boolean => {
+  const { stakingAssetId } = new ChainRegistryClient().bundled.globals();
+  let feeAssetId = txv?.bodyView?.transactionParameters?.fee?.assetId;
+  if (feeAssetId === undefined) {
+    feeAssetId = stakingAssetId;
+  }
+
+  return feeAssetId.equals(stakingAssetId);
+};
+
 export const TransactionApproval = () => {
   const { authorizeRequest: authReqString, setChoice, sendResponse } = useStore(txApprovalSelector);
 
   const { selectedTransactionView, selectedTransactionViewName, setSelectedTransactionViewName } =
     useTransactionViewSwitcher();
-
-  const [symbol, setSymbol] = useState<string>('Unknown asset');
 
   if (!authReqString) {
     return null;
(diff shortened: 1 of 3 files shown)


hdevalence (changes requested):
The warning actually looks great, but let's put it at the top of the window rather than the bottom.

apps/extension/src/routes/popup/approval/transaction/index.tsx
```diff
@@ -51,6 +51,17 @@ export const TransactionApproval = () => {
...
+          <div
+            style={{ marginBottom: '16px' }}
+            className='rounded border border-yellow-500 p-2 text-yellow-500 text-sm'
+          >
+            <span className='block text-center font-bold'>⚠ Privacy Warning:</span>
+            Transaction uses a non-native fee token. To reduce gas costs and protect your privacy,
+            maintain an UM balance for fees.
+          </div>
+        )}
+
```
TalDerei, line 64: alternatively, we can implement a tooltip with a hover animation

Screenshot 2024-11-25 at 12 31 47 AM

hdevalence: I think it's fine, hopefully the user only sees this once or twice

hdevalence (approved):

apps/extension/src/routes/popup/approval/transaction/index.tsx
```diff
@@ -47,7 +50,18 @@ export const TransactionApproval = () => {
 
   return (
     <div className='flex h-screen flex-col'>
-      <div className='grow overflow-auto p-[30px] pt-10'>
+      <div className='flex grow flex-col overflow-auto p-[30px] pt-10'>
+        {selectedTransactionViewName === TransactionViewTab.SENDER && symbol !== 'UM' && (
```
grod220, line 54: suggestion: think this would be simpler if we ditched the whole useEffect pattern and just did something like:

```ts
const hasAltGasFee = (txv?: TransactionView): boolean => {
  const { stakingAssetId } = new ChainRegistryClient().bundled.globals();
  return txv?.bodyView?.transactionParameters?.fee?.assetId?.equals(stakingAssetId) ?? false;
};
...

  return (
    <div className='flex h-screen flex-col'>
      <div className='flex grow flex-col overflow-auto p-[30px] pt-10'>
        {selectedTransactionViewName === TransactionViewTab.SENDER &&
          hasAltGasFee(selectedTransactionView) && (
            <div
                 ...
```

apps/extension/src/routes/popup/approval/transaction/index.tsx
```diff
@@ -47,7 +50,18 @@ export const TransactionApproval = () => {
...
+          <div
+            style={{ marginBottom: '16px' }}
```
grod220, line 56: issue: `className="mb-4"`

apps/extension/src/routes/popup/approval/transaction/index.tsx
TalDerei: helper function is a better approach here 👍

TalDerei:
```ts

const hasAltGasFee = (txv?: TransactionView): boolean => {
  const { stakingAssetId } = new ChainRegistryClient().bundled.globals();
  
  let feeAssetId = txv?.bodyView?.transactionParameters?.fee?.assetId;
  if (feeAssetId === undefined) {
    feeAssetId = stakingAssetId;
  }

  return stakingAssetId.equals(stakingAssetId) ?? false;
};
```
...

```ts
{selectedTransactionViewName === TransactionViewTab.SENDER && !hasAltGasFee(selectedTransactionView) && (
```

apps/extension/src/routes/popup/approval/transaction/index.tsx
TalDerei: moved to classname

TalDerei:
I’ll wait to merge until https://github.com/prax-wallet/prax/pull/241 is complete, so I can compare how the privacy warning appears in the updated approval dialog design.

grod220 (approved):
Much simpler 👍

PR #248, opened 2024-12-03
Correct import paths

grod220:
A follow up from https://github.com/prax-wallet/prax/pull/247

Legacy build methods causes the typescript compiler to create a bunch of compiled files locally. Also, fixes import paths that were throwing eslint errors after making the `lint` script the same as the other packages.


TalDerei (approved):
tested locally 👍

PR #254, opened 2024-12-18
privacy warning: t-addr

TalDerei:
adds a privacy warning to the approval dialog if the `Ics20Withdrawal` message includes a transparent address (t-addr).

CI pipeline is currently failing because this depends on https://github.com/penumbra-zone/web/pull/1950 and requires bumping package dependencies.

-------------

Screenshot 2024-12-18 at 11 52 55 AM


TalDerei:
need to truncate IBC-address rendering which is currently getting cut-off

----------------

Screenshot 2024-12-18 at 1 17 47 PM

hdevalence:
Shouldn't the updated rendering show "use transparent address" rather than "use compat address"?

apps/extension/src/routes/popup/approval/transaction/index.tsx
```diff
@@ -65,11 +75,18 @@ export const TransactionApproval = () => {
...
-              Transaction uses a non-native fee token. To reduce gas costs and protect your privacy,
-              maintain an UM balance for fees.
+              {hasTransparentAddress(selectedTransactionView) ? (
+                <p>This transaction uses transparent addresses which may reduce privacy.</p>
+              ) : (
+                <p>
+                  Transaction uses a non-native fee token. To reduce gas costs and protect your
+                  privacy, maintain an UM balance for fees
+                </p>
+              )}
```
hdevalence, line 89: Why is this an either/or? Shouldn't we have two separate warnings, since each case can happen independently? If they both occur they can be stacked on top of each other.

TalDerei: sure, we can stack these privacy warnings

TalDerei: I wasn't sure if we wanted to display both privacy warnings, so `t-addr` had a higher priority.

TalDerei: Screenshot 2024-12-20 at 3 12 49 PM

TalDerei: Screenshot 2024-12-20 at 3 19 24 PM

TalDerei:
the approval dialogue needs to be updated to reflect the `t-addr` for USDC Noble withdrawls. 

Screenshot 2024-12-20 at 3 44 05 PM

TalDerei:
> Shouldn't the updated rendering show "use transparent address" rather than "use compat address"?

yes, updating the approval dialogue to reflect the `t-addr` for USDC Noble withdrawals.

erwanor:
I am not going to block on it, but if we set the transparent flag on, I think it makes sense to render the withdrawal component with a return address encoded as a transparent address.

We do this by adding a `PENUMBRA_BECH32_TRANSPARENT_ADDRESS_PREFIX` with helpers to the web package `@penumbra-zone/bech32`. Then, we should be able to display the transparent encoding of an `Address`. This is how it's done in core: https://github.com/penumbra-zone/penumbra/blob/020986747245270a00d039306458be9e7f1f33f5/crates/core/keys/src/address.rs#L232

erwanor (approved):
LGTM, on green CI, with requested web changes (other pr) that should make it easy to do taddr rendering in the withdrawal component

TalDerei:
> We do this by adding a `PENUMBRA_BECH32_TRANSPARENT_ADDRESS_PREFIX` with helpers to the web package `@penumbra-zone/bech32`

Screenshot 2024-12-21 at 10 39 35 PM

Issue #230, opened 2024-11-07
Genesis block diagnosis


vacekj:
**Issue:** Sync takes forever for the genesis block on airdrop accounts. Specific to low-power devices.
This doesn't impact non-airdrop accounts, even on slow devices.

**Steps to diagnose:**
1. Make a local devnet with a new airdrop account, or use a mainnet airdrop account
2. Make a VM with low memory and CPU, or use a low-powered physical device.
4. Install prax onto a fresh browser profile
5. Observe the initial genesis sync getting stuck/going very slowly.
6. Profile the JS and WASM to find out the issue.

PR #257, opened 2024-12-23
feat: chunk genesis sync

TalDerei:
references https://github.com/prax-wallet/prax/issues/230. CI is currently failing because it depends on function definitions in https://github.com/penumbra-zone/web/pull/1975. cc @vacekj 

In https://github.com/prax-wallet/prax/pull/165, we bundled the mainnet genesis state with other extension assets. This PR handles syncing the bundled genesis block more manageably with a chunking approach, which also improves extension popup loading times during onboarding for _existing wallets_. Longer-term, the optimal solution is to migrate syncing out of the service worker and into a dedicated worker thread, freeing up the service worker. 

Currently, the service worker initializes the [wallet services ](https://github.com/prax-wallet/prax/blob/main/apps/extension/src/service-worker.ts#L47) and invokes the block processor interface to sync the genesis block. Until now, lower-powered devices and slower internet connections struggled to handle this heavy processing task, which presented some undesirable side-effects:

**1. Blocking onboarding:** extension's connection popup delays from appearing promptly after onboarding, approximately **30s+** on public wifi on my machine. The delay occurs because genesis sync requires crossing the wasm-boundary and performing heavy cryptographic operations on ~32K state payloads in the compact block.  Consequently, the block processor couldn’t yield to the event loop to finish setting up the listeners responsible for initializing the content scripts while performing this operation, leaving the service worker continuously busy with genesis block sync.

**2. Failed connection requests:** a long-running genesis sync fails to acquire a connection handle. 

----------------

_before_

Screenshot 2024-12-22 at 5 29 09 PM

https://github.com/user-attachments/assets/f225771d-dfd0-433a-91af-03be6e41144d

----------------

_after_

https://github.com/user-attachments/assets/d63db302-daee-4b30-9540-efdbb3a34e21


TalDerei:
currently failing with SCT corruption error.

TalDerei:
note to reviewers: putting on hold for now in favor of dex-explorer work

erwanor:
+1 let's put this on hold for now

TalDerei:
update: @turbocrime had a great idea that maybe we can generalize this beyond just the genesis block to any congested block with a large state payload set.

TalDerei:
> update: @turbocrime had a great idea that maybe we can generalize this beyond just the genesis block to any congested block with a large state payload set.

on the other hand, the maximum payload for a compact block is 100KB (think the max we've since in practice is half of that density). SNRs are ~700 bytes, so roughly ~140 txs / block max. the genesis block is a special case with ~32K state payloads.

erwanor (approved):
This received empirical testing, so I'll ACK optimistically.

TalDerei:
simplified the implementation by removing extraneous helper methods in favor of calling the wasm functions directly, and reusing the existing `processBlock` method after handling the genesis block. This reduces the number of assumptions made post-chunking.

PR #263, opened 2025-01-10
Prep v15.1 release

TalDerei:
deps need to be updated to incorporate the major version bump for wasm introduced in https://github.com/penumbra-zone/web/pull/1953. 

pairs with https://github.com/penumbra-zone/dex-explorer/pull/236.

_update:_ will treat this as a staging area for when we decide to cut next release

PR #269, opened 2025-01-20
feat: idle syncing

TalDerei:
uses [chrome.alarms](https://developer.chrome.com/docs/extensions/reference/api/alarms) API to restart the service worker after extended periods of inactivity and make sync progress over time without any frontend providers or the extension being opened. This ensures the chain is already synced when the user returns to penumbra after that extended period of not using it. We aptly name this '**idle syncing**'. 

- accompanying context doc: https://docs.google.com/document/d/18awgAeCIhwVLmfKj_R7pXoFTNtcOMtdtKhCWRs0iDgE/edit?tab=t.0
(this a long winded sanity check for sidestepping a major refactor to our connection lifecycle)

cc @hdevalence 

-------------------

when adding an event listener to log details as the alarm triggers (set `periodInMinutes` to 1 minute) and inspecting the service worker immediately after manually terminating it with `chrome.runtime.reload()`, it appears to working at first glance. However, we should contemplate potential pitfalls with this approach. 

Screenshot 2025-01-20 at 3 19 24 PM

see https://github.com/prax-wallet/prax/pull/269#issuecomment-2608598293 for an example

### DIFF
--- a/apps/extension/src/service-worker.ts
+++ b/apps/extension/src/service-worker.ts
@@ -96,3 +96,9 @@ const handler = await backOff(() => initHandler(), {
 });
 
 CRSessionManager.init(PRAX, handler);
+
+// https://developer.chrome.com/docs/extensions/reference/api/alarms
+chrome.alarms.create('blockSync', {
+  periodInMinutes: 30,
+  delayInMinutes: 0
+});
--- a/apps/extension/src/service-worker.ts
+++ b/apps/extension/src/service-worker.ts
@@ -102,3 +102,9 @@ void chrome.alarms.create('blockSync', {
   periodInMinutes: 30,
   delayInMinutes: 0,
 });
+
+chrome.alarms.onAlarm.addListener(alarm => {
+  if (alarm.name === 'blockSync') {
+    console.log('Block sync alarm fired, service worker restarting');
+  }
+});
--- a/apps/extension/src/service-worker.ts
+++ b/apps/extension/src/service-worker.ts
@@ -105,6 +105,8 @@ void chrome.alarms.create('blockSync', {
 
 chrome.alarms.onAlarm.addListener(alarm => {
   if (alarm.name === 'blockSync') {
-    console.log('Block sync alarm fired, service worker restarting');
+    if (globalThis.__DEV__) {
+      console.info('Background sync scheduled');
+    }
   }
 });
--- a/apps/extension/src/service-worker.ts
+++ b/apps/extension/src/service-worker.ts
@@ -98,7 +98,7 @@ const handler = await backOff(() => initHandler(), {
 CRSessionManager.init(PRAX, handler);
 
 // https://developer.chrome.com/docs/extensions/reference/api/alarms
-chrome.alarms.create('blockSync', {
+void chrome.alarms.create('blockSync', {
   periodInMinutes: 30,
-  delayInMinutes: 0
+  delayInMinutes: 0,
 });
--- a/apps/extension/public/manifest.json
+++ b/apps/extension/public/manifest.json
@@ -40,7 +40,7 @@
       "matches": ["<all_urls>"]
     }
   ],
-  "permissions": ["storage", "unlimitedStorage", "offscreen"],
+  "permissions": ["storage", "unlimitedStorage", "offscreen", "alarms"],
   "host_permissions": ["<all_urls>"],
   "externally_connectable": {
     "matches": ["<all_urls>"]


apps/extension/src/service-worker.ts
```diff
@@ -96,3 +96,9 @@ const handler = await backOff(() => initHandler(), {
 });
 
 CRSessionManager.init(PRAX, handler);
+
+// https://developer.chrome.com/docs/extensions/reference/api/alarms
+void chrome.alarms.create('blockSync', {
+  periodInMinutes: 30,
```
TalDerei, line 102: dynamically adjustable param

hdevalence:
One test that comes to mind is, with each version of the code (pre and post change):

1. ensure no frontends or any other tabs are open 
2. clear cache to reset view state
3. immediately quit chrome (to kill the service worker)
4. start chrome, do not open any tabs or frontends, wait 1h
5. open minifront
6. observe the current sync progress on minifront opening

(a custom chrome profile may be helpful here)

If background syncing is working we expect the post-change test to have significantly higher sync height at (6). Exactly how much is less important — as long as it's increasing we have reason to believe it will catch up and stay up to date.

TalDerei:
per the [docs](https://developer.chrome.com/docs/extensions/reference/api/alarms): "starting in Chrome 117, the number of active alarms is limited to 500. Once this limit is reached, chrome.alarms.create() will fail. When using a callback, [chrome.runtime.lastError](https://developer.chrome.com/docs/extensions/reference/api/runtime#property-lastError) will be set. When using promises, the promise will be rejected."

seems like a necessary extension to this might be to periodically clear all alarms using `chrome.alarms.clearAll()` in a listener.

TalDerei:
> One test that comes to mind is, with each version of the code (pre and post change):

going through this flow and setting the `periodInMinutes == 1 minute` synced the entire chain on minifront opening.

apps/extension/src/service-worker.ts
erwanor: FWIW, I think it will have to be much shorter than this (between 20 seconds and 4 minutes)

https://developer.chrome.com/docs/extensions/develop/concepts/service-workers/lifecycle

Screenshot 2025-01-21 at 7 52 48 AM

TalDerei: > FWIW, I think it will have to be much shorter than this (between 20 seconds and 4 minutes)

"starting in Chrome 120, the minimum alarm interval has been reduced from 1 minute to 30 seconds"

I've been testing locally with 1 minute.

TalDerei:
hey!!

> i considered alarms unnecessary because every content script (meaning every page load) will attempt to message the extension background with an init message, to connect pre-authorized origins. this message should wake up the extension background.

...

> it should be possible to refactor this in a way that avoids alarms and any new alarm permission, by refactoring [the init message handler](https://github.com/prax-wallet/prax/blob/main/apps/extension/src/listeners/content-script/init.ts) to require the return value of `CRSessionManager.init` to be in scope, or to execute `CRSessionManager.init` with a similar technique. the simplest way to do that would just mean relocating the content of `listeners/content-script/init.ts` into `service-worker.ts` so a top-level value can be captured in its callback's closure.

it's true that the `CRSessionClient` is used by [content scripts](https://github.com/prax-wallet/prax/blob/main/apps/extension/src/content-scripts/injected-connection-port.ts#L17) injected by the extension that run in the context of the web page, and the content scripts initiate a connection to the manager, which causes chrome to queue the connection request while the service worker initializes.

but, the use case we're targeting is when the user doesn't open the extension and load a frontend for prolonged periods, which wouldn't trigger any content script init messages, and thus make no sync progress while the user's been offline. 

> the simplest way to handle this with alarms is, create an alarm with `periodInMinutes`, and a listener, in `service-worker.ts`. call `CRSessionManager.init` in the callback, and leave the current call where it is, so it also starts up immediately without waiting for the alarm.

thought the same thing initially, but in practice just creating the alarm is sufficient to wake up the service worker - seems we don't need to put `CRSessionManager.init` inside a listener.

> `CRSessionManager` maintains a global singleton state that should prevent creation of multiple session managers. you should confirm this works the way i expect. if it doesn't, maybe you can assign the return value of `CRSessionManager.init` to a top-level variable.

yeah, CRSessionManager's [singleton pattern](https://github.com/penumbra-zone/web/blob/main/packages/transport-chrome/src/session-manager.ts#L44) prevents multiple session managers, so subsequent calls just return the existing [sessions map](https://github.com/penumbra-zone/web/blob/main/packages/transport-chrome/src/session-manager.ts#L44).

---------------------

 alarms are a nice idiomatic solution for periodically waking up the service worker. I do wonder however how the additional permissions will impact chrome store reviews.

turbocrime:
an alarm is definitely the idiomatic way to do it

TalDerei:
> One test that comes to mind is, with each version of the code (pre and post change):

@hdevalence replicated the procedure on my older 2020 m1 pro – the browser cache was cleared, and neither any frontends nor the extension were opened.

https://github.com/user-attachments/assets/94ec7c4a-a436-4ef2-bf61-e8f5830841bc

after starting chrome and waiting 10 minutes, it synced the entire chain. Previously, it would have just blocked on genesis sync during service worker init without making any progress. 

https://github.com/user-attachments/assets/e5ad629e-080b-48f6-b9d3-7e677eb3fc6c

currently experimenting with the suggestion @turbocrime left above.

hdevalence:
Well, if the alarm approach solves the sync problem, that seems like a good improvement, and we could continue to refine it over time. A 30 minute alarm seems low frequency but we can find out if there are objections.

TalDerei:
> Well, if the alarm approach solves the sync problem, that seems like a good improvement, and we could continue to refine it over time. A 30 minute alarm seems low frequency but we can find out if there are objections.

an overnight test with a 30 minute frequency successfully kept up with the chain, and it was fully synced when I checked this morning.

Screenshot 2025-01-23 at 7 36 31 AM

TalDerei:
note to reviewers: this will be staged in https://github.com/prax-wallet/prax/pull/263 and cut as part of the next `v15.1` next prax release.

apps/extension/src/service-worker.ts
```diff
@@ -102,3 +102,9 @@ void chrome.alarms.create('blockSync', {
   periodInMinutes: 30,
   delayInMinutes: 0,
 });
+
+chrome.alarms.onAlarm.addListener(alarm => {
+  if (alarm.name === 'blockSync') {
+    console.log('Block sync alarm fired, service worker restarting');
```
TalDerei, line 108: for checking beta reviewer logs

erwanor: This shows up in the service worker console right? not the main web one

TalDerei: logs display in both since chrome forwards the service worker logs to the main extension console

erwanor (approved):
Cool, good work. I think we should get this in the hands of beta testers asap. The wake up frequency seems polite and reasonable so I hope Chrome review won't object.

PR #273, opened 2025-01-24
docs: ci/cd creds for ext publishing

conorsch:
Jotting down some notes while memory is fresh.


conorsch:
I don't understand the linting failure. Would appreciate if someone else did the needful and pulled this in. Low-priority, since this is essentially docs for posterity.

TalDerei (approved):

PR #279, opened 2025-01-28
feat(query): update block processor to include subaccounts into LPs

VanishMax:
Relates to https://github.com/penumbra-zone/web/pull/2006 (Web) and https://github.com/penumbra-zone/dex-explorer/pull/329 (DEX)

To run this change:
1. Run `pnpm dev:pack` PR in https://github.com/penumbra-zone/web/pull/2006
2. In Prax, run `pnpm add:tgz $PATH_TO_WEB_REPO/packages/*/*.tgz`
3. Build the extension and clear the cache, so it re-starts the sync

To test:
- Wait for the sync to complete
- Open up the DEX PR with updated `protobuf` package and run it locally
- Check if the position filter works

This PR must be merged after the packages from Web are published and applied here.


packages/query/src/helpers/identify-txs.ts
```diff
@@ -147,28 +159,48 @@ const generateTxId = async (tx: Transaction): Promise<TransactionId> => {
...
-  spentNullifiers: Set<Nullifier>,
+  spentNullifiers: Map<Nullifier, SpendableNoteRecord | SwapRecord>,
   commitmentRecords: Map<StateCommitment, SpendableNoteRecord | SwapRecord>,
-  isControlledAddr: ViewServerInterface['isControlledAddress'],
+  getIndexByAddress: ViewServerInterface['getIndexByAddress'],
 ): Promise<
   { relevantTx: RelevantTx; recoveredSourceRecords: RecoveredSourceRecords } | undefined
 > => {
   let txId: TransactionId | undefined; // If set, that means this tx is relevant and should be returned to the caller
+  let subaccount: AddressIndex | undefined;
```
VanishMax, line 185: this change collects subaccounts from `SpendableNoteRecord`s but there's no way to get AddressIndex for all possible actions, that's why it might be `undefined`. However, it is enough to get them for the purpose of the tasks – to identify LPs per subaccount

vacekj (approved):
lgtm.

packages/query/src/block-processor.ts
```diff
@@ -494,7 +494,7 @@ export class BlockProcessor implements BlockProcessorInterface {
 
   // Nullifier is published in network when a note is spent or swap is claimed.
   private async resolveNullifiers(nullifiers: Nullifier[], height: bigint) {
-    const spentNullifiers = new Set<Nullifier>();
+    const spentNullifiers = new Map<Nullifier, SpendableNoteRecord | SwapRecord>();
```
erwanor, line 497: Before we merge, I want to circle back to this and a few other things with the data modeling. What's the lifetime of the spent nullifier map? How big is a SNR? An OOM seems somewhat unlikely but I don't totally understand the approach yet.

erwanor (changes requested):
I think this is a great start, and I want to test it.

packages/query/src/block-processor.ts
erwanor: OK this seems totally fine, round it to 1KB, that's 1M SNRs. sgtm

TalDerei: after reviewing the code, `Map<Nullifier, SpendableNoteRecord | SwapRecord>` is actually the optimal and clever approach to _minimize_ db fetches required to understand what subaccount a particular position is associated with. 

stepping through the code, prior to calling `resolveNullifiers()` , indexdb stores the successfully trial decrypted spendable notes in a table who's primary key is the commitment and secondary index is the nullifier. So in theory, you have all the information to know which SNR is associated with what commitment and nullifier directly in storage, but don't _yet_ know which transaction that commitment is associated with, because we haven't processed and saved the transaction information at this stage of the block processor. 

`resolveNullifiers()` will subsequently [query](wallet/prax/pull/279/files#diff-ccb0493c986b958cf859e773170457e5ff59a4a2e0d742698492247f6ae872daR497) any SNR / SwapRecord stored in indexeddb that match any nullifier in the incoming compact block, and the record <> nullifier pair are stored in the [spentNullifiers](https://github.com/prax-wallet/prax/pull/279/files#diff-ccb0493c986b958cf859e773170457e5ff59a4a2e0d742698492247f6ae872daR497) map. 

up until this point in the block processing, we know a few things:
- which SNR / SwapRecord is associated with the user
- what commitment + nullifier is associated with what SNR / SwapRecord
- don't know which transaction is associated with the SNR / SwapRecord or commitment

then, if we found a commitment or nullifier relevant to the user in that block, we request all the transactions for that block height from the full node, and [filter the specific transaction](https://github.com/prax-wallet/prax/pull/279/files#diff-df9a3308a5ae26e8f5e912c4bdd659a76d12e6f751ec7c3eb8de3c0b81147bf0R234) relevant to the user. This is where the original `spentNullifiers` map comes into play. We have a `spentNullifiers` map (Map<Nullifier, SpendableNoteRecord | SwapRecord>) and `commitmentRecords` map (Map<StateCommitment, SpendableNoteRecord | SwapRecord>), and once we know which commitment / nullifier inside the transaction's actions is associated with the user's commitment / nullifier (stored in those maps in memory), we know we transaction is the user's, and we return the transaction associated with it's relevant subacount by deriving the `AddressIndex` from the SNR itself. 

After identifying the transaction, we have a special method called [processTransactions](https://github.com/prax-wallet/prax/pull/279/files#diff-ccb0493c986b958cf859e773170457e5ff59a4a2e0d742698492247f6ae872daR550-R556) that will identify the LP position relevant to the user and save it alongside the subaccount. 

anyways, if we didn't originally have this map, once we knew which nullifier is associated with the user, we'd need to perform another db query to fetch the relevant SNR from the table, and subsequently derive the AddressIndex.

packages/query/src/helpers/identify-txs.ts
```diff
@@ -16,78 +16,89 @@ import { FungibleTokenPacketData } from '@penumbra-zone/protobuf/penumbra/core/c
...
  * Identifies if a tx has a relay action of which the receiver is the user.
+ * If so, returns corresponding address.
+ *
  * In terms of minting notes in the shielded pool, three IBC actions are relevant:
  * - MsgRecvPacket (containing an ICS20 inbound transfer)
  * - MsgAcknowledgement (containing an error acknowledgement, thus triggering a refund on our end)
  * - MsgTimeout
  */
-const hasRelevantIbcRelay = (
+const getRelevantIbcRelay = (
```
TalDerei, line 34: comment: see comments above about possibly not modifying this since we don't have a use case for it yet.

packages/query/src/helpers/identify-txs.ts
```diff
@@ -147,28 +159,48 @@ const generateTxId = async (tx: Transaction): Promise<TransactionId> => {
...
-  isControlledAddr: ViewServerInterface['isControlledAddress'],
+  getIndexByAddress: ViewServerInterface['getIndexByAddress'],
 ): Promise<
   { relevantTx: RelevantTx; recoveredSourceRecords: RecoveredSourceRecords } | undefined
 > => {
   let txId: TransactionId | undefined; // If set, that means this tx is relevant and should be returned to the caller
+  let subaccount: AddressIndex | undefined;
   const recoveredSourceRecords: RecoveredSourceRecords = [];
 
+  // for spend/swapClaim transaction actions
```
TalDerei, line 188: suggestion: can we please add better code comments here for processing nullifiers + commitments?

packages/query/src/helpers/identify-txs.ts
```diff
@@ -147,28 +159,48 @@ const generateTxId = async (tx: Transaction): Promise<TransactionId> => {
...
+      subaccount = getAddressIndexFromNote(spendableNoteRecord);
     }
   }
 
+  // For output/swap/swapClaim transaction actions
   const txCommitments = getCommitmentsFromActions(tx);
   for (const [stateCommitment, spendableNoteRecord] of commitmentRecords) {
     if (txCommitments.some(txCommitment => stateCommitment.equals(txCommitment))) {
       txId ??= await generateTxId(tx);
+      subaccount = getAddressIndexFromNote(spendableNoteRecord);
```
TalDerei, line 203: comment: minor optimization for the future; If `getCommitmentsFromActions` is successful, we can skip `getNullifiersFromActions` and `relevantIbcRelay` because they calculate the same txId and subaccount fields. 

we can address that separately when we conduct more in-depth block processor performance optimizations. I’ll open a separate issue for it.

TalDerei (approved):
this is really great work! left some feedback and good to merge once we bump the proto deps and green CI

packages/query/src/helpers/identify-txs.ts
TalDerei: https://github.com/prax-wallet/prax/issues/280

packages/query/src/block-processor.ts
VanishMax: thanks for this great explanation, @TalDerei! This is exactly what the PR does in the smallest details

packages/query/src/block-processor.ts
```diff
@@ -317,7 +317,7 @@ export class BlockProcessor implements BlockProcessorInterface {
         spentNullifiers,
         recordsByCommitment,
         blockTx,
-        addr => this.viewServer.isControlledAddress(addr),
+        addr => this.viewServer.getIndexByAddress(addr),
```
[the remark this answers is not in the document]
VanishMax: put back the `isControlledAddress` method. Truly, no need to know AddressIndex for IBC relay actions for now

erwanor (approved):

Issue #175, opened 2024-08-23
remove minifront workaround code


turbocrime:
https://github.com/prax-wallet/web/pull/173 is a workaround that should be removed when a new pd ships with a new bundled frontend that does not experience the issue.

i'm filing this issue before #173 merges so i can link an issue URL from a comment in that PR.

TalDerei:
latest pd https://github.com/penumbra-zone/penumbra/releases/tag/v0.80.3

turbocrime:
this was reverted at some point

https://github.com/prax-wallet/prax/blob/7bcf54de3ba9bf07576ddd0561d8143f0d10d437/apps/extension/src/content-scripts/injected-penumbra-global.ts#L60-L68

PR #285, opened 2025-02-04
don't retain port in content scripts

turbocrime:
Content scripts no longer retain a `MessagePort` in violation of object transfer rules.

The `injected-penumbra-global.ts` content script has no insight to channel state, and as a mainworld script should not be relied upon to manage channel state. So this takes the approach of always requesting a new port.

The `injected-listeners.ts` content script cannot re-provide a port, since it must retain no reference. So this depends on changes in https://github.com/penumbra-zone/web/pull/2018 so it may launch a new session for each request.


apps/extension/public/manifest.json
```diff
@@ -17,8 +18,9 @@
   "content_scripts": [
     {
       "matches": ["https://*/*", "http://localhost/*"],
-      "js": ["injected-connection-port.js", "injected-request-listener.js"],
-      "run_at": "document_start"
+      "js": ["injected-listeners.js"],
```
turbocrime, line 21: consolidated listener scripts

apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -19,7 +19,8 @@
 
 import '@penumbra-zone/client/global';
 
-import { createPenumbraStateEvent, type PenumbraProvider } from '@penumbra-zone/client';
+import { createPenumbraStateEvent } from '@penumbra-zone/client/event';
+import type { PenumbraProvider } from '@penumbra-zone/client/provider';
```
turbocrime, line 23: not strictly necessary, but due to poor webpack tree-shaking, importing from the root package apparently caused the generated script to be very large.

a more specific import significantly reduces the size of the content script.

apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -51,25 +52,14 @@ class PraxInjection {
...
   private injection: Readonly<PenumbraProvider> = Object.freeze({
-    /**
-     * Meet the 'request' method of the old page API to mitigate incompatibility
-     * with pd v0.80.0's bundled minifront. This prevents connection failure.
-     * @todo Remove when bundled frontends are updated beyond `a31d54a`
-     * @issue https://github.com/prax-wallet/web/issues/175
-     */
-    request: async () => {
-      await Promise.resolve(this.port ?? this.postConnectRequest());
-    },
```
turbocrime, line 68: re-fixes #175 which was reverted by https://github.com/prax-wallet/prax/pull/236

maybe this isn't appropriate for this PR - should it remain?

apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -51,25 +52,14 @@ class PraxInjection {
...
-
-    connect: () => Promise.resolve(this.port ?? this.postConnectRequest()),
+    connect: () => this.postConnectRequest(),
     disconnect: () => this.postDisconnectRequest(),
-    isConnected: () => Boolean(this.port && this.presentState === PenumbraState.Connected),
+    isConnected: () => this.presentState === PenumbraState.Connected,
```
turbocrime, line 62: i would like to do some more investigation to identify possibly unknown problems with these changes to state transitions

apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -108,10 +95,19 @@ class PraxInjection {
     return attempt;
   }
 
+  private postDisconnectRequest() {
+    const attempt = this.listenEndMessage();
+    window.postMessage(disconnectMessage, '/', []);
+    return attempt;
+  }
```
turbocrime, line 102: pulled this out into its own method to match the structure of the connect request logic

apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -108,10 +95,19 @@ class PraxInjection {
...
+
   private listenPortMessage() {
-    this.setPending();
+    if (this.presentState !== PenumbraState.Connected) {
+      this.setPending();
+    }
```
turbocrime, line 107: this change needs some closer inspection too

turbocrime (commented):
commentary. possibly should minimize extra changes

apps/extension/src/content-scripts/injected-penumbra-global.ts
turbocrime: reverted

TalDerei: I removed this again in https://github.com/prax-wallet/prax/pull/213, but it was reverted in https://github.com/prax-wallet/prax/pull/236 as you pointed out. Is there a way to determine if we can remove this temporary workaround and the issue has been fully resolved?

TalDerei (commented):
looks like the changeset primarily consolidated content scripts, removed global port object, and better message handling.

TalDerei (approved):
changed base branch to `connection-lifecycle-refactor`

apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -57,6 +57,14 @@ class PraxInjection {
...
 
   private injection: Readonly<PenumbraProvider> = Object.freeze({
+    /**
+     * Meet the 'request' method of the old page API to mitigate incompatibility
+     * with pd v0.80.0's bundled minifront. This prevents connection failure.
+     * @todo Remove when bundled frontends are updated beyond `a31d54a`
+     * @issue https://github.com/prax-wallet/web/issues/175
+     */
+    request: () => this.postConnectRequest(),
+
```
TalDerei, line 67: let's handle this in separate PR as discussed

PR #292, opened 2025-02-12
prax: LQT integration 

TalDerei:
## Description of Changes

CI pipeline is currently failing because this depends on https://github.com/penumbra-zone/web/pull/2041 and requires bumping package dependencies. easier to review the diff by unchecking 'show annotations'. 

end-to-end testing across view service, block processor, wasm, and indexedDB is challenging across separate PRs. This serves as a staging area to test and build on the code drafted in those individual draft PRs. 

references tracking issue: https://github.com/penumbra-zone/web/issues/2022 and pairs with https://github.com/penumbra-zone/web/pull/2041.

----------------------

**View Service:**

- extends queriers to the funding service, enabling nullifier checks for voting eligibility against against rpc hosted by pd.

-------------

**Block Processor:**

- Implements the necessary block processor plumbing to detect reward notes during sync based on the `LiquidityTournamentReward` commitment source variant. This enables prax to track LQT votes and reward notes and save them into the `LqtHistoricalVotes` table.


packages/query/src/helpers/identify-txs.ts
```diff
@@ -178,10 +196,24 @@ const searchRelevant = async (
...
+  // then rehydrated the relevant transaction sources for the SNRs and returned the relevant transaction.
+  //
+  // We checked nullifiers in [spend, swapClaim] actions and state commitments in [output, swap, swapClaim] actions.
+  // However, if you inspect the intersection of these actions, you'll notice that spends will always be associated
+  // with outputs (expect for spend), and swapClaims contain both nullifiers and commitments. Therefore, we don't need
+  // to check nullifiers? But in fact, we need nullifiers because what if for instance we sent our entire balance to
+  // another external account, which will only produce spends (and not outputs) relevent to us.
+  //
+  // We only need to rehydrate the transaction sources for SNRs, as the other source variants (eg. LQT) are already populated.
+
```
TalDerei, line 216: observations from spending time here determining if we needed to rehydrate sources for LQT-derived SNRs.

cronokirby (commented):
So I think the attribution of rewards isn't correctly handled (see comments in review).

Also, one thing I haven't quite gotten yet from both prs: do we query an RPC for figuring out spent LQT nullifiers, or our local database? If local, I think we should also, in that same method, mark those nullifiers as being voted with in that epoch.

packages/query/src/block-processor.ts
```diff
@@ -577,6 +581,46 @@ export class BlockProcessor implements BlockProcessorInterface {
     }
   }
 
+  /**
+   * Identify liquidity tournament votes and rewards.
+   */
+  private async identifyLiquidityTournamentVotes(
```
cronokirby, line 587: I don't think this function correctly identifies the reward associated with this particular vote. My understanding of the logic as it is currently is that we:
- check if the action is an LQT vote,
- and that we had spend authority for the note at some point in time,
- and if that note had as its source the lqt, then mark that note as the reward.

But, what we want to identify as the reward is not the note being spent to *vote* in the LQT, but instead the note that gets minted at the end of the epoch *because* of that vote. For example, there can't be any reward after we've voted until the end of an epoch, but this logic will identify reward notes from epoch N used to vote in epoch N + 1 as being the rewards for the current epoch. I think instead, to associate rewards, we should use the fact that the LQT commitment source links to a transaction id, and use that to update our votes with that tx id with the reward.

TalDerei:
> Also, one thing I haven't quite gotten yet from both prs: do we query an RPC for figuring out spent LQT nullifiers, or our local database? If local, I think we should also, in that same method, mark those nullifiers as being voted with in that epoch.

that happens [here](https://github.com/penumbra-zone/web/pull/2041/files#diff-ef2eb249051b31b8540e77cdeb74cf12db4c52c1b68eb8cc2b8b2bb302bc6e04R28-R40) in the `lqtVotingNotes` view service impl. 

we request all delegation notes from our [local database](https://github.com/penumbra-zone/web/blob/main/packages/storage/src/indexed-db/index.ts#L586-L588) that haven't been spent yet, and then make nullifier queries against the [funding service](https://github.com/prax-wallet/prax/pull/292/files#diff-4d81d4b4e0479139026fc9cc21ee49c428eca4d65373675f8ae7e277833ed938R20-R38), which is backed by a querier implemented in prax. The querier returns only the subset of delegation notes that haven't been used for voting in the current epoch and haven’t been marked as spent. If it doesn't return any delegation notes, the wasm planner will [throw accordingly](https://github.com/penumbra-zone/web/pull/2041/files#diff-ec51d55a0b0ab2ce263357c3d43509ddfe1d88ace8bc963213a0899cca79a0f0R666-R670). 

these view service requests are subsequently constructed in the frontend requesting this information.

packages/query/src/block-processor.ts
TalDerei: good catch, the intention was _not_ to use the note being spent to vote as the reward note, but that's the current behavior because I _accidentally_ wrapped the entire code logic in that if statement. 

tracing what's happening here against pcli, I have 2 delegation notes; 1 delegation note and 1 delegation note (from previous epoch) in block N, and then in block N + 1, it will cycle through the identifyLiquidityTournamentVotes for each action. It will overwrite my vote with the seemingly correct reward (because the reward for each epoch is the same right now as I'm the only one voting), and that's what I didn't spot it. The key outlier here that proves this is incorrect is it has the wrong tx id. 


-------------
Screenshot 2025-02-14 at 2 01 31 PM


Screenshot 2025-02-14 at 2 01 20 PM

TalDerei: good catch, the intention was _not_ to use the note being spent to vote as the reward note, but that's the current behavior because I _accidentally_ wrapped the entire code logic in that if statement. 

tracing what's happening here against pcli, I have 2 delegation notes – one from the current epoch and one from the previous epoch in block N (as a reward), and then in block N + 1, it cycles through `identifyLiquidityTournamentVotes` for each action. Since there are two liquidity tournament votes in the transaction, the second action overwrites my vote with what appears to be the correct reward (since the reward for each epoch is currently the same as I’m the only one voting), which is why I didn’t notice it initially. The key outlier here that proves that this is incorrect is it has the wrong tx id for the reward. 

I'll tighten the logic in this method, which felt like the shakiest part of this PR. 

----------------

Screenshot 2025-02-14 at 2 01 31 PM

Screenshot 2025-02-14 at 2 01 20 PM

TalDerei: fixed the logic by separating the voting and reward detection for clarity, which I probably the approach I should've taken from the start.

packages/query/src/block-processor.ts
```diff
@@ -582,39 +600,81 @@ export class BlockProcessor implements BlockProcessorInterface {
...
-      const incentivizedTokenAssetId = new AssetId({
-        altBaseDenom: action.value.body?.incentivized?.denom,
-      });
+    let totalVoteWeight: Amount | undefined = undefined;
+    let incentivizedAssetMetadata: Metadata | undefined = undefined;
 
-      const incentivizedAssetMetadata =
-        await this.indexedDb.getAssetsMetadata(incentivizedTokenAssetId);
+    for (const { action } of transaction.body?.actions ?? []) {
+      let currentVoteWeight: Amount | undefined;
```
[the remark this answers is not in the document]
TalDerei: Interesting, thought they were summable, and the sum of the delegation tokens was your voting weight. but you're saying you can't because delegation tokens can represent stakes in different validators and may have different exchange rates?  

I can easily modify it to track votes by asset id instead of storing a cumulative value.

packages/ui/components/ui/tx/actions-views/liquidity-tournament-vote.tsx
```diff
@@ -0,0 +1,38 @@
...
+            )}
+          </ActionDetails>
+        }
+      />
+    );
+  }
+
+  if (value.liquidityTournamentVote.case === 'opaque') {
+    return <ViewBox label='Liquidity Tournament Vote' />;
+  }
```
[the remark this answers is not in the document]
TalDerei: new actions views are underway in https://github.com/penumbra-zone/web/issues/2053, and doesn't make sense to modify our deprecated UI library here actually. I'll punt on the actions views for now in anticipation of the implementation in our new UI library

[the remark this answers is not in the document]
TalDerei: added address view in visible case, and existing fields in opaque case. again, still lacking, but sufficient for testing purposes

packages/query/src/block-processor.ts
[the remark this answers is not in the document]
TalDerei: refactored to display multiple votes per epoch, segmented by the corresponding delegation token's asset id. The LQT table maintains a unique arbitrary primary key and indexes entries by epoch.

--------

Screenshot 2025-02-19 at 11 41 45 AM

TalDerei:
@cronokirby bumped to latest released npm deps

cronokirby (approved):

PR #294, opened 2025-02-17
shorten countdown timer in prax

TalDerei:
references https://github.com/penumbra-zone/web/issues/2042

-----------

https://github.com/user-attachments/assets/10fa5420-4d41-46b9-b7dc-54f0c156cb2b


hdevalence:
There's still a noticeable delay, could we eliminate it entirely?

TalDerei:
> There's still a noticeable delay, could we eliminate it entirely?

https://github.com/user-attachments/assets/abdd14a5-790f-42c6-8510-cc15e8df6ef5

PR #296, opened 2025-02-19
fix `getChainId` startup request

### LINKED ISSUE #277: sometimes unable to reconnect to extension after connection is lost
losing connection breaks the injected content scripts and listeners they are supposed to initialize, preventing reconnecting again for some timeout period defined in our transport layer. To my knowledge, this the reconnection period wasn't always that long, pointing to a potential regression somewhere?

--------------
These observations are from commits **prior** to `0a6edc7` (_idle syncing_ feature)

1. If the connection is lost and neither the webpage or extension is open, the webpage can successfully reload and allow syncing to resume. 

https://github.com/user-attachments/assets/a1432f78-c570-4448-b36d-75b4fa01be61

-----------

2. If the connection is lost in the webpage, and the webpage is reloaded when there isn't an active internet connection, it prevents reconnecting again. 

https://github.com/user-attachments/assets/d0f7f91c-bbf1-40a5-8367-a86475f81f35

--------------------

3. If the connection is lost in the webpage, and the extension is opened when there isn't an active internet connection, it prevents reconnecting again. 

https://github.com/user-attachments/assets/5798a99e-be7b-4242-9419-31081086dd09

------------

4. If the extension remains open during while the connection is lost, it will recover gracefully after the internet connection is established again. 

https://github.com/user-attachments/assets/0105da93-20be-414e-97e7-a192547a0740

turbocrime:
fixes #277

### DIFF
--- a/apps/extension/src/wallet-services.ts
+++ b/apps/extension/src/wallet-services.ts
@@ -35,13 +35,18 @@ export const startWalletServices = async () => {
 };
 
 /**
- * Get the chain id from local storage, or the rpc endpoint if no chain id is in
- * local storage.
+ * Get chainId from the rpc endpoint, or fall back to chainId from storage.
+ *
+ * It's possible that the remote endpoint may suddenly serve a new chainId.
+ * @see https://github.com/prax-wallet/prax/pull/65
  */
 const getChainId = async (baseUrl: string) => {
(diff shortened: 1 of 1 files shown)


TalDerei (approved):
fixes the failure cases in https://github.com/prax-wallet/prax/issues/277

PR #297, opened 2025-02-19
Connection lifecycle work

turbocrime:
Staging branch for improving reliability of connection lifecycle.

Session reconnect requires manager init change after https://github.com/penumbra-zone/web/pull/2034

Relevant to https://github.com/penumbra-zone/web/pull/2063

Possibly pending https://github.com/penumbra-zone/web/pull/2062


apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -147,14 +160,20 @@ class PraxInjection {
...
     return disconnection.promise;
   }
+
+  private ambientEndListener = (msg: MessageEvent<unknown>) => {
+    if (msg.origin === window.origin) {
+      if (isPraxEndMessageEvent(msg)) {
+        this.setDisconnected();
+      }
+    }
+  };
```
turbocrime, line 176: new listener

apps/extension/src/senders/revoke.ts
turbocrime, line 1: new end message sent on kill

TalDerei (approved):
logic is largely the same since we last reviewed it together, looks good to merge!

PR #303, opened 2025-03-06
Build cycle improvements

turbocrime:
don't fetch packages during tgz install. this is nice for offline devleopment, as the build iteration doesn't spend time waiting on the network.

syncpack no longer cares about sort location of pnpm field in package json, which add:tgz creates in the 'wrong' place

webpack now only launches chrome when in watch mode, so if a CHROMIUM_PROFILE var is set ambiently, it's not always trying to open chrome.

### DIFF
--- a/.syncpackrc
+++ b/.syncpackrc
@@ -12,8 +12,7 @@
     "scripts",
     "files",
     "exports",
-    "publishConfig",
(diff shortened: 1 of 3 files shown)


TalDerei (approved):
LGTM

PR #304, opened 2025-03-07
prax: minimal t-addr reversion 

TalDerei:
requires bumping package deps to pass CI

corollary to https://github.com/penumbra-zone/web/issues/2023 and sparsely pairs with https://github.com/penumbra-zone/web/pull/2101

### DIFF
--- a/apps/extension/src/state/tx-approval.ts
+++ b/apps/extension/src/state/tx-approval.ts
@@ -97,7 +97,7 @@ export const createTxApprovalSlice = (): SliceCreator<TxApprovalSlice> => (set, 
       state.txApproval.asSender = asSender.toJsonString();
       state.txApproval.asPublic = asPublic.toJsonString();
       state.txApproval.asReceiver = asReceiver.toJsonString();
-      state.txApproval.transactionClassification = transactionClassification;
+      state.txApproval.transactionClassification = transactionClassification.type;
 
       state.txApproval.choice = undefined;
     });
--- a/packages/ui/components/ui/tx/actions-views/isc20-withdrawal.tsx
+++ b/packages/ui/components/ui/tx/actions-views/isc20-withdrawal.tsx
@@ -48,10 +48,6 @@ export const Ics20WithdrawalComponent = ({ value }: { value: Ics20Withdrawal }) 
             </ActionDetails.Row>
           )}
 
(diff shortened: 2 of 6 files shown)


packages/ui/components/ui/tx/actions-views/isc20-withdrawal.tsx
```diff
@@ -48,10 +48,6 @@ export const Ics20WithdrawalComponent = ({ value }: { value: Ics20Withdrawal })
             </ActionDetails.Row>
           )}
 
-          <ActionDetails.Row label='Use Transparent Address'>
-            {value.useTransparentAddress ? 'TRUE' : 'FALSE'}
-          </ActionDetails.Row>
-
```
TalDerei, line 54: cosmetic change based on feedback

apps/extension/src/state/tx-approval.ts
```diff
@@ -97,7 +97,7 @@ export const createTxApprovalSlice = (): SliceCreator<TxApprovalSlice> => (set,
       state.txApproval.asSender = asSender.toJsonString();
       state.txApproval.asPublic = asPublic.toJsonString();
       state.txApproval.asReceiver = asReceiver.toJsonString();
-      state.txApproval.transactionClassification = transactionClassification;
+      state.txApproval.transactionClassification = transactionClassification.type;
```
cronokirby, line 100: I think this line is causing the lint errors, and I'm not sure sure what the purpose of it is?

cronokirby (changes requested):
I think we should try and fix the lints related to the change in tx-approval, and I'm not quite sure what the purpose of it is.

apps/extension/src/state/tx-approval.ts
TalDerei: > I think this line is causing the lint errors, and I'm not sure sure what the purpose of it is?

https://github.com/penumbra-zone/web/pull/2083/files#diff-70eac8f3879c22852e9eb3caec7e267983300a42216b24b2b8a130623a346e6fR10

requires bumping the perspective package dep

TalDerei: forgot the npm packages were already updated, bumping now

cronokirby (approved):

PR #307, opened 2025-03-13
refactor: internal messaging

turbocrime:
This is a draft PR to refactor the internal messaging system.

### DIFF
--- a/apps/extension/src/entry/popup-root.tsx
+++ b/apps/extension/src/entry/popup-root.tsx
@@ -1,36 +1,33 @@
+import { ConnectError } from '@connectrpc/connect';
+import { errorToJson } from '@connectrpc/connect/protocol-connect';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { StrictMode, useState } from 'react';
 import { createRoot } from 'react-dom/client';
 import { RouterProvider } from 'react-router-dom';
+import { isPopupRequest, PopupType } from '../message/popup';
 import { popupRouter } from '../routes/popup/router';
-
-import { isOriginApprovalRequest, isPopupRequest, isTxApprovalRequest } from '../message/popup';
 import { useStore } from '../state';
 import { originApprovalSelector } from '../state/origin-approval';
 import { txApprovalSelector } from '../state/tx-approval';
-
-import { errorToJson } from '@connectrpc/connect/protocol-connect';
-import { ConnectError } from '@connectrpc/connect';
 
 import '@repo/ui/styles/globals.css';
 
 chrome.runtime.onMessage.addListener(
   (req: unknown, _: chrome.runtime.MessageSender, responder: (x: unknown) => void) => {
     if (isPopupRequest(req)) {
-      try {
-        if (isTxApprovalRequest(req)) {
-          void txApprovalSelector(useStore.getState()).acceptRequest(req, responder);
-        } else if (isOriginApprovalRequest(req)) {
-          originApprovalSelector(useStore.getState()).acceptRequest(req, responder);
-        } else {
-          throw new Error('Unknown popup request');
+      void Promise.resolve(req).then(async req => {
+        try {
+          if (isPopupRequest(req, PopupType.TxApproval)) {
+            responder(await txApprovalSelector(useStore.getState()).acceptRequest(req));
+          } else if (isPopupRequest(req, PopupType.OriginApproval)) {
+            responder(await originApprovalSelector(useStore.getState()).acceptRequest(req));
+          } else {
+            throw new TypeError('Unknown popup request', { cause: req });
+          }
+        } catch (e) {
+          responder({ error: errorToJson(ConnectError.from(e), undefined) });
         }
-      } catch (e) {
-        responder({
-          type: req.type,
-          error: errorToJson(ConnectError.from(e), undefined),
-        });
-      }
+      });
       return true; // instruct chrome runtime to wait for a response
     }
     return false; // instruct chrome runtime we will not respond
--- a/apps/extension/src/message/popup.ts
+++ b/apps/extension/src/message/popup.ts
@@ -1,49 +1,54 @@
 import type { AuthorizeRequest } from '@penumbra-zone/protobuf/penumbra/custody/v1/custody_pb';
-import type {
-  InternalMessage,
-  InternalRequest,
-  InternalResponse,
-} from '@penumbra-zone/types/internal-msg/shared';
-import type { UserChoice } from '@penumbra-zone/types/user-choice';
 import type { Jsonified } from '@penumbra-zone/types/jsonified';
+import type { UserChoice } from '@penumbra-zone/types/user-choice';
 import { OriginRecord } from '../storage/types';
+import { JsonValue } from '@bufbuild/protobuf';
 
 export enum PopupType {
   TxApproval = 'TxApproval',
   OriginApproval = 'OriginApproval',
 }
 
-export type PopupMessage = TxApproval | OriginApproval;
-export type PopupRequest<T extends PopupMessage = PopupMessage> = InternalRequest<T>;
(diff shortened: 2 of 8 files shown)


TalDerei (commented):
haven't reviewed in-depth yet, but we should generally be mindful of avoiding _excessive_ diffs. It's not entirely clear what specific issue this addresses in the our popup state management list or if it potentially introduces unnecessary scope creep? I'll spend some more time going through the changeset though.

turbocrime:
> haven't reviewed in-depth yet, but we should generally be mindful of avoiding _excessive_ diffs. It's not entirely clear what specific issue this addresses in the our popup state management list or if it potentially introduces unnecessary scope creep? I'll spend some more time going through the changeset though.

i'm presently breaking this PR down into a series of more conceptually-scoped changes, similar in approach  to the recent work on transports. will PR some of that progress in a moment.

apps/extension/src/entry/popup-root.tsx
turbocrime, line 1: initialization is now handled by a hook

apps/extension/src/hooks/popup-init.ts
turbocrime, line 1: this could be cleaned up a bit, but it brings all the dialog ready/init operations to one location, with tighter async state management

apps/extension/src/message/popup.ts
turbocrime, line 1: eliminates use of `@penumbra-zone/types/internal-msg` for popup messaging

apps/extension/src/message/popup.ts
```diff
@@ -1,49 +1,65 @@
...
-export type PopupRequest<T extends PopupMessage = PopupMessage> = InternalRequest<T>;
-export type PopupResponse<T extends PopupMessage = PopupMessage> = InternalResponse<T>;
+export type PopupError = Record<'error', JsonValue>;
 
-export type OriginApproval = InternalMessage<
-  PopupType.OriginApproval,
-  { origin: string; favIconUrl?: string; title?: string; lastRequest?: number },
-  null | OriginRecord
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type PopupRequest<M extends PopupType = any> = Record<
```
turbocrime, line 15: working with these types gets a lot more difficult without this explicit 'any'

in practice, the type is parameterized

apps/extension/src/state/origin-approval.ts
turbocrime, line 1: slices are refactored to return promises for results instead of directly using the message responder

apps/extension/src/state/tx-approval.ts
```diff
@@ -41,122 +39,127 @@ export interface TxApprovalSlice {
...
-    const getMetadata = async (assetId: AssetId) => {
-      try {
-        const { denomMetadata } = await viewClient.assetMetadataById({ assetId });
-        return denomMetadata ?? new Metadata({ penumbraAssetId: assetId });
-      } catch {
-        return new Metadata({ penumbraAssetId: assetId });
-      }
-    };
-
-    const wallets = await localExtStorage.get('wallets');
```
turbocrime, line 71: this `localExtStorage` access is now refactored into a state initializer parameter, like all the other slices' storage access. this was necessary to implement tests.

turbocrime (commented):
minimized changes out of the draft and staging branches. i think a more comprehensive refactor is still beneficial, but this is effective and smaller

turbocrime:
since last review: types are improved, handling is improved.

instead of attempting to track existence of popups across documents by inspecting chrome extension context (which creates race conditions), the web locks API is used to create a global extension lock for the popup type.

apps/extension/src/popup.ts
```diff
@@ -1,114 +1,103 @@
...
-      if (msg === popupId) {
-        resolve();
-        chrome.runtime.onMessage.removeListener(idListen);
-        respond();
-      }
-    };
-
-    chrome.runtime.onMessage.addListener(idListen);
+  await chrome.windows.create({
+    url: popupUrl(popupType, id).href,
```
TalDerei, line 97: comment: ACK – unique embedded url with unique popup Id

apps/extension/src/message/listen-popup.ts
```diff
@@ -0,0 +1,24 @@
...
+    popupId: string,
+    handle: <T extends PopupType>(message: PopupRequest<T>) => Promise<PopupResponse<T>>,
+  ) =>
+  (
+    message: unknown,
+    sender: chrome.runtime.MessageSender,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    sendResponse: (response: PopupResponse<any> | PopupError) => void,
+  ): boolean => {
+    if (isInternalSender(sender) && isPopupRequest(popupId, message)) {
```
TalDerei, line 17: comment: ACK – enforces that message is intended for this specific popup, by matching the popup’s id.  avoids popup windows potentially receiving and responding to messages not meant for it

apps/extension/src/popup.ts
```diff
@@ -1,114 +1,103 @@
...
-const POPUP_READY_TIMEOUT = 60 * 1000;
+/**
+ * Spawns a popup with a unique id, and resolves the ID when the popup is ready.
+ * Ready promise times out in {@link POPUP_READY_TIMEOUT} milliseconds.
+ */
+const spawnPopup = async (popupType: PopupType): Promise<string> => {
+  const id = crypto.randomUUID();
+  const ready = listenReady(id, AbortSignal.timeout(POPUP_READY_TIMEOUT));
+
+  const geometry = await chrome.windows
```
TalDerei, line 86: comment: not entirely sure what the geometry logic is doing here or why it differs from the previous implementation?

TalDerei (approved):
really great work 👍 modularity helped reason about the logic

apps/extension/src/popup.ts
```diff
@@ -1,114 +1,103 @@
...
+    const popupRequest = {
+      [popupType]: request,
+      id: await spawnPopup(popupType),
+    } as PopupRequest<M>;
+
+    return sendPopup(popupRequest);
+  }) satisfies LockGrantedCallback;
+
+  const popupResponse: PopupResponse<M> | null = await (navigator.locks.request(
+    popupType,
```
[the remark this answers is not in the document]
TalDerei: > the `popupType` parameter is restricted to only permit members of `PopupType`

let me rephrase: "wondering if there are contrived scenarios here where the lock would need to be enforced on `popup` directly rather than `popupType`?" for example, could it be possible to simulate triggering multiple different popup types simultaneously?

I realize that's a different question.

apps/extension/src/popup.ts
```diff
@@ -1,114 +1,103 @@
...
-    });
-
-  if (response && 'error' in response) {
-    throw errorFromJson(response.error, undefined, ConnectError.from(response));
+import { sessionExtStorage } from './storage/session';
+import { PopupRequest, PopupResponse, PopupType } from './message/popup';
+import { sendPopup } from './message/send-popup';
+import { listenReady } from './message/listen-ready';
+
+const POPUP_READY_TIMEOUT = 60_000;
```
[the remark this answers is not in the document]
TalDerei: yeah this doesn't actually seem like a big deal to me, since the uncommon-case here would be extremely slow-loading popups. agree to punt on this for now

apps/extension/src/popup.ts
TalDerei: really helpful context; so if I'm understanding correctly: In macos, offscreen geometry is sanitized by the OS, but in windows / linux, chrome can basically open a popup in a _nearly_ invisible location (which these geometry changes partially mitigate). 

wondering why we haven't encountered user complaints about this before? one naive assumption is that there aren't many windows / linux users using penumbra in the first place. 

also, why can't we access the `window.screen` global directly in the popup scripts (ie. react UI in `popup.html`)? how do other chrome extensions handle this?

Issue #177, opened 2024-08-24
Implement better and more consistent incoming message validation in all message handlers


turbocrime:
Message listeners perform inconsistent validation of incoming messages, and test coverage is poor.

different listeners may validate input with

- type guards
- helper functions
- string literal comparison
- an enumerated switch
- object inspection code directly in the handler

Type guards tend to be colocated with message type definitions, so are not obviously critical to message validation in listeners. These should be documented and called out with docstrings.

Object inspection code inside handlers should be converted to type guards with docstrings, and relocated with the message types.

Some existing helper methods could be applied more broadly.

Some validation functions have test coverage, but presently, all listeners and type guards have zero test coverage. These paths are security critical and should have full test coverage, with special attention to:

- malformed input
- input that satisfies multiple types
- listeners expecting input from specific sources

PR #309, opened 2025-03-18
refactor: content script listener

turbocrime:
Use `sendTab` and `sendTabs` wrappers for `chrome.tabs.sendMessage`

makes progress on https://github.com/prax-wallet/prax/issues/177


apps/extension/src/listeners/content-script/disconnect.ts
```diff
@@ -1,33 +1,41 @@
...
+      revokeOrigin(targetOrigin);
+      respond(null);
     }
-    revokeOrigin(validSender.origin);
-    void chrome.tabs.sendMessage(validSender.tab.id, PraxConnection.End, {
-      // end only the specific document
-      frameId: validSender.frameId,
-      documentId: validSender.documentId,
-    });
   });
```
turbocrime, line 38: this behavior is changed to be more careful about communicating with pages that should not have permission

apps/extension/src/listeners/content-script/disconnect.ts
```diff
@@ -1,33 +1,41 @@
...
-      throw new Error('Sender does not possess approval');
+      if (globalThis.__DEV__) {
+        console.warn('Sender requesting disconnect does not possess approval', validSender);
+      }
+      // this is strange, but session termination still makes sense
+      CRSessionManager.killOrigin(targetOrigin);
+      respond(PenumbraRequestFailure.Denied);
+    } else {
+      revokeOrigin(targetOrigin);
+      respond(null);
```
TalDerei, line 36: comment: can you explain the session termination in this code block?

apps/extension/src/listeners/external/message-external.ts
```diff
@@ -1,6 +0,0 @@
-// this is temporary code to use the externally_connectable permission,
-// also providing an easter egg for curious users
-export const praxEasterEgg: ChromeExtensionMessageEventListener = (_, __, response) => {
```
TalDerei, line 3: comment: should we remove `externally_connectable` from manifest then?

turbocrime: this should be reverted

apps/extension/src/listeners/content-script/disconnect.ts
turbocrime: storage check is the worst operation here and might be worth preventing.

apps/extension/src/listeners/external/message-external.ts
TalDerei: I didn't mean to publish that comment actually, I was thinking about something else here. I was wondering why the easter egg was removed.

TalDerei (approved):
good to merge after reverting easter egg

PR #310, opened 2025-03-18
refactor: relocate listeners

turbocrime:
begins to collect the extension worker's listeners in a single location, to encourage consistency and clarify listener lifetimes.

this is a minor change but if merged into a previous or later stage, the diff would become unclear.


TalDerei (commented):
honestly seems extraneous to me, don't think this is strictly necessary. The only thing we should probably do here is move the content-script and internal message listeners  into the service worker.

turbocrime:
> honestly seems extraneous to me, don't think this is strictly necessary.

it's not a functional change.

it makes unit testing of the listeners possible.

breaking out this specific change also reduced the diff in a subsequent PR introducing wrappers, which was closed to reduce dependent scope pending progress here

TalDerei:
> it makes unit testing of the listeners possible.

sure, can we stick with the original file structure for the listeners? https://github.com/prax-wallet/prax/tree/main/apps/extension/src/listeners

turbocrime:
> can we eventually stick with something resembling the original folder organization for the listeners? https://github.com/prax-wallet/prax/tree/main/apps/extension/src/listeners

that's https://github.com/prax-wallet/prax/tree/99e4b9222be0294483acf23fe5cdc3cf2acedc5d/apps/extension/src/message/listen

PR #311, opened 2025-03-18
refactor: introduce `ControlMessage`

turbocrime:
introduce a local type for internal prax messaging.

update utilities to use the new types, with equivalent functionality.


turbocrime:
reducing scope

PR #314, opened 2025-03-25
refactor: popup slices use promises to respond

turbocrime:
instead of handling message responders directly, popup slices simply provide a promise for their result. the responder function remains within the message listener scope.

tests are added to confirm functionality in the slices, then the change is implemented.

### DIFF
--- a/apps/extension/src/state/origin-approval.ts
+++ b/apps/extension/src/state/origin-approval.ts
@@ -1,12 +1,10 @@
-import { ConnectError } from '@connectrpc/connect';
 import { OriginApproval, PopupType } from '../message/popup';
 import { AllSlices, SliceCreator } from '.';
-import { errorToJson } from '@connectrpc/connect/protocol-connect';
 import type { InternalRequest, InternalResponse } from '@penumbra-zone/types/internal-msg/shared';
 import { UserChoice } from '@penumbra-zone/types/user-choice';
 
 export interface OriginApprovalSlice {
-  responder?: (m: InternalResponse<OriginApproval>) => void;
+  responder?: PromiseWithResolvers<InternalResponse<OriginApproval>>;
   favIconUrl?: string;
   title?: string;
   requestOrigin?: string;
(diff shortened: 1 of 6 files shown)


apps/extension/src/state/tx-approval.ts
```diff
@@ -40,123 +39,126 @@ export interface TxApprovalSlice {
...
+    setChoice: choice => {
+      set(state => {
+        state.txApproval.choice = choice;
       });
-    }
-  },
-});
+    },
+
+    sendResponse: () => {
```
TalDerei, line 115: comment: In what context may the promise remain unresolved? If I decide to close the extension approval dialogue or refresh page rather than approving / denying, will the slice still call `sendResponse`? It's unclear whether we're currently handling these cases, which could lead to hanging promises.

TalDerei:
@turbocrime will do another round of async review after all PRs have merged into this staging area, thanks for breaking them apart

apps/extension/src/state/tx-approval.ts
turbocrime: that behavior is unchanged by this PR.

https://github.com/prax-wallet/prax/blob/edba81b497ec3013fac3c26cb449303f8b286044/apps/extension/src/popup.ts#L28-L36

apps/extension/src/state/origin-approval.ts
```diff
@@ -30,48 +27,51 @@ export const createOriginApprovalSlice = (): SliceCreator<OriginApprovalSlice> =
...
+        responder.resolve({
+          type: PopupType.OriginApproval,
+          data: {
+            choice,
+            origin: requestOrigin,
+            date: Date.now(),
+          },
+        });
+      } catch (e) {
+        responder.reject(e);
```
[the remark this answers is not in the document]
turbocrime: error serialization is performed where the popup's listener catches the rejection.

https://github.com/prax-wallet/prax/blob/edba81b497ec3013fac3c26cb449303f8b286044/apps/extension/src/entry/popup-root.tsx#L33-L38

apps/extension/src/state/tx-approval.ts
TalDerei: nice that makes sense

turbocrime: refreshing the popup destroys the slice, and it won't be reconfigured. apparently chrome fails to recognize the resoponder is dropped when the popup is refreshed.

added a listener to close the popup on navigation, including refresh.

PR #322, opened 2025-04-14
fix(ui): #2235: approval dialog styles

VanishMax:
Closes https://github.com/penumbra-zone/web/issues/2235

Small changes for better display of action views in the old designs of the action views.

After the changes:

image

image

### DIFF
--- a/.changeset/happy-needles-warn.md
+++ b/.changeset/happy-needles-warn.md
@@ -1,0 +1,5 @@
+---
+'@repo/ui': patch
+---
+
+Fix styles of action views in approval dialog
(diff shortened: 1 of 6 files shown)


TalDerei (approved):
henry had functional requirements for accounts to display on the same line, but these overflow scenarios weren’t accounted for which this fixes appropriately. 

the current prax redesign was implemented somewhat ad hoc (https://github.com/prax-wallet/prax/pull/241), and we’ll likely want to prioritize a _proper_ redesign and updating the UI components down the line.

PR #324, opened 2025-04-14
process epoch consistently

turbocrime:
depends on 

- [ ] https://github.com/penumbra-zone/web/pull/2242
- [ ] https://github.com/penumbra-zone/web/pull/2258

the intent here was to move epoch increment before flush, because `saveScanResult` called here

https://github.com/prax-wallet/prax/blob/a61f49bcbc0929d8d491357a5b1ebed38d677af9/packages/query/src/block-processor.ts#L282-L290

calls `addNewSwaps` https://github.com/penumbra-zone/web/blob/51eea2bcd83cfc16c8afe5afe9a69b637cfe523f/packages/storage/src/indexed-db/index.ts#L222-L232

calls `getEpochByHeight` https://github.com/penumbra-zone/web/blob/51eea2bcd83cfc16c8afe5afe9a69b637cfe523f/packages/storage/src/indexed-db/index.ts#L1000-L1010

which breaks if `getEpochByHeight` is fixed, as in  https://github.com/penumbra-zone/web/pull/2242

so now, epoch update is handled more consistently.

the validator state update could be reverted partially or improved, but validator state update is also sensitive to changes in epoch change.

### DIFF
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -227,7 +227,9 @@ export class BlockProcessor implements BlockProcessorInterface {
         currentEpoch++;
         await this.indexedDb.addEpoch({ index: currentEpoch, startHeight: currentHeight });
 
-        await this.indexedDb.updateValidatorInfo(this.querier.stake.allValidatorInfos());
+        if (compactBlock.height === 0n || latestKnownBlockHeight === compactBlock.height) {
+          await this.indexedDb.updateValidatorInfo(this.querier.stake.allValidatorInfos());
+        }
       }
     }
   }
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -231,6 +231,10 @@ export class BlockProcessor implements BlockProcessorInterface {
           await this.indexedDb.updateValidatorInfo(this.querier.stake.allValidatorInfos());
         }
       }
+
+      if (globalThis.__ASSERT_ROOT__) {
+        await this.assertRootValid(compactBlock.height);
+      }
     }
   }
 
@@ -384,10 +388,6 @@ export class BlockProcessor implements BlockProcessorInterface {
         compactBlock.swapOutputs,
         compactBlock.height,
       );
-    }
-
-    if (globalThis.__ASSERT_ROOT__) {
-      await this.assertRootValid(compactBlock.height);
     }
   }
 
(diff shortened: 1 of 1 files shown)


packages/query/src/block-processor.ts
```diff
@@ -167,78 +193,53 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+    let currentEpoch =
+      currentHeight && (await this.indexedDb.getEpochByHeight(currentHeight)).index;
 
-        await this.processBlock({
-          compactBlock: this.genesisBlock,
-          latestKnownBlockHeight: latestKnownBlockHeight,
-          skipTrialDecrypt,
-        });
-      }
-    }
```
TalDerei, line 197: comment: special-casing genesis sync is something we want to preserve; the genesis block already carries unique assumptions in our block processing logic and should be segmented accordingly. By encapsulating this logic, rather than embedding genesis checks directly into the block processing method, we make functional improvements like https://github.com/prax-wallet/prax/pull/257 easier to implement and reason about. This is outside the scope of the issue.

packages/query/src/block-processor.ts
```diff
@@ -167,78 +193,53 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+    currentEpoch ??= PRE_GENESIS;
 
     // this is an indefinite stream of the (compact) chain from the network
     // intended to run continuously
-    for await (const compactBlock of this.querier.compactBlock.compactBlockRange({
-      startHeight: currentHeight + 1n,
-      keepAlive: true,
-      abortSignal: this.abortController.signal,
-    })) {
+    for await (const compactBlock of this.streamBlocks(currentHeight)) {
```
TalDerei, line 205: comment: remove genesis special-casing from async block querier helper

packages/query/src/block-processor.ts
```diff
@@ -167,78 +193,53 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+      if (compactBlock.height > latestKnownBlockHeight) {
+        latestKnownBlockHeight = compactBlock.height;
+      }
```
TalDerei, line 208: comment: what's the reason for moving this check before block processing? The behavior appears functionally the same – let's revert it and why wouldn't we keep the contextual comment about tendermint.

packages/query/src/block-processor.ts
```diff
@@ -167,78 +193,53 @@ export class BlockProcessor implements BlockProcessorInterface {
...
       // confirm block height to prevent corruption of local state
-      if (compactBlock.height === currentHeight + 1n) {
-        currentHeight = compactBlock.height;
-      } else {
+      if (compactBlock.height !== currentHeight) {
```
TalDerei, line 210: comment: previous conditional logic `compactBlock.height === currentHeight + 1n` seems more precise

packages/query/src/block-processor.ts
```diff
@@ -167,78 +193,53 @@ export class BlockProcessor implements BlockProcessorInterface {
...
 
-      await this.processBlock({
-        compactBlock: compactBlock,
-        latestKnownBlockHeight: latestKnownBlockHeight,
-        skipTrialDecrypt,
-      });
+      if (compactBlock.height === 0n) {
+        void this.indexedDb.updateValidatorInfo(this.querier.stake.allValidatorInfos());
+        // await this.handleGenesis(compactBlock);
+      }
```
TalDerei, line 221: comment: this can move back into special-casing the genesis block

packages/query/src/block-processor.ts
```diff
@@ -475,28 +471,6 @@ export class BlockProcessor implements BlockProcessorInterface {
     return undefined;
   }
 
-  private async saveAndReturnDelegationMetadata(
```
TalDerei, line 478: comment: revert validator methods

packages/query/src/block-processor.ts
```diff
@@ -167,78 +193,53 @@ export class BlockProcessor implements BlockProcessorInterface {
...
-  private async processBlock({
-    compactBlock,
-    latestKnownBlockHeight,
-    skipTrialDecrypt = false,
-  }: ProcessBlockParams) {
+  private async processBlock({ compactBlock, latestKnownBlockHeight }: ProcessBlockParams) {
+    const skipTrialDecrypt = shouldSkipTrialDecrypt(
+      this.walletCreationBlockHeight,
+      compactBlock.height,
+    );
```
TalDerei, line 242: comment: ACK – moving `skipTrialDecrypt` into `processBlock`. we can even revert to reduce diff because it's not strictly necessary.

packages/query/src/block-processor.ts
```diff
@@ -167,78 +193,53 @@ export class BlockProcessor implements BlockProcessorInterface {
...
-        currentHeight = this.genesisBlock.height;
-
-        // Set the trial decryption flag for the genesis compact block
-        const skipTrialDecrypt = shouldSkipTrialDecrypt(
-          this.walletCreationBlockHeight,
-          currentHeight,
-        );
+    let currentHeight = await this.indexedDb.getFullSyncHeight();
+    let currentEpoch =
+      currentHeight && (await this.indexedDb.getEpochByHeight(currentHeight)).index;
```
TalDerei, line 198: comment: keep `currentEpoch` and revert `currentHeight` based on subsequent comments

packages/query/src/block-processor.ts
```diff
@@ -167,78 +193,53 @@ export class BlockProcessor implements BlockProcessorInterface {
...
-        latestKnownBlockHeight = compactBlock.height;
+      await this.processBlock({ compactBlock, latestKnownBlockHeight });
+
+      currentHeight++;
+      if (compactBlock.epochRoot) {
+        currentEpoch++;
+        await this.indexedDb.addEpoch({ index: currentEpoch, startHeight: currentHeight });
+
+        if (compactBlock.height === 0n || latestKnownBlockHeight === compactBlock.height) {
+          await this.indexedDb.updateValidatorInfo(this.querier.stake.allValidatorInfos());
```
TalDerei, line 231: comment: this is the meat of the proposed fix, ie. epoch tracking and the atomic handling of epoch transitions further down; everything else is extraneous towards that

packages/query/src/block-processor.ts
turbocrime: the genesis block does not require unique assumptions in the block processing logic. treating the genesis block as a special case makes it more difficult to ensure block processing applies consistent logic to every block. treating a bundled genesis block as a special-special case makes it additionally difficult.

when the genesis block is not bundled, for example on testnets and devnets or any other chain with no bundled genesis block, this special-case logic does not activate. so, the block processor's normal iteration must be able to process a genesis block anyway. so you have three conditions, instead of no conditions.

the chunking logic applied in #257 should be applied generally if it is truly an improvement. if it only sees benefit on blocks with large state payload, consider the size of the state payload the condition, instead of the block height, and ensure it is applied in a manner that is valid for any block with a large state payload.

i expect to step back from this PR to focus on existing work items, so change as you think necessary.

packages/query/src/block-processor.ts
turbocrime: marking resolved, this will follow solution to https://github.com/prax-wallet/prax/pull/324#discussion_r2047167522

packages/query/src/block-processor.ts
TalDerei: I can take over this PR

packages/query/src/block-processor.ts
turbocrime: the `latestKnownBlockHeight` parameter provided to `processBlock` was formerly used in epoch logic inside `processBlock`, which was part of the refactor concerning epoch issues.

you're right that changing this behavior could affect other things. a quick look indicates one of the `flushReasons` checks becomes constant-false, so this needs some kind of fix.


when updates are flushed, the block height in the flush is written to the database, and other calls within `processBlock` will ultimately read the database value, instead of the value passed as a parameter here. so, in those cases, operations are performed with a latest known block height that is not equal to this value, unless the value is updated at the top of the call.

so for correctness, i think this should remain updated at the top of the iteration, or should not be passed to `processBlock` at all. this will provide better/complete consistency with operations using the database check.

if nothing else, conceptually, a block arriving with a new height should change the latest known block height. 

probably, the checks within `processBlock` that actually use the `latestKnownBlockHeight` should expect that the current block height is not greater.

packages/query/src/block-processor.ts
turbocrime: marking resolved, this will follow solution to https://github.com/prax-wallet/prax/pull/324#discussion_r2047167522

packages/query/src/block-processor.ts
turbocrime: the increment as implemented here is correct, but if

- https://github.com/prax-wallet/prax/pull/324#discussion_r2047205170
- https://github.com/prax-wallet/prax/pull/324#discussion_r2047182849

are changed then this logic is of course no longer valid.

TalDerei:
closing in favor of https://github.com/prax-wallet/prax/pull/345

PR #326, opened 2025-04-16
remove minifront workaround

TalDerei:
references https://github.com/prax-wallet/prax/issues/175


turbocrime (approved):
approving, nothing should still rely on this

PR #320, opened 2025-04-09
ledger support

turbocrime:
depends on https://github.com/penumbra-zone/web/pull/2229/

successfully signs transactions requested by frontend.

ui is not amazing, onboarding may be fragile.

required more significant storage refactor than i expected, due to many assumptions about the availability of a seed phrase. i want to split this out into phases.

PR #333, opened 2025-04-28
DRAFT: migrate storage for multiple wallet types

turbocrime:
draft PR. work is coming from https://github.com/prax-wallet/prax/pull/320 but trying to break it down into phases.

migration also requires changes to all locations that use the changed structures, so it's tangled.


TalDerei:
@turbocrime we should reprioritize this migration and flush out unit test coverage, then manual testing can proceed.

TalDerei:
@turbocrime what's the status of the deconstruction here towards the migration?

PR #336, opened 2025-05-01
feat(quick-sync): TCT frontier support

TalDerei:
pairs with https://github.com/penumbra-zone/web/pull/2303

currently getting connectrpc errors when trying to query the `sctFrontier` rpc method because full node on testnet hasn't picked up the latest SCT service.

### DIFF
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -171,6 +171,10 @@ export class BlockProcessor implements BlockProcessorInterface {
     // prepares for syncing and checks for a bundled genesis block,
     // which can save time by avoiding an initial network request.
     if (currentHeight === PRE_GENESIS_SYNC_HEIGHT) {
+
+      // STUB SCT CALL
+      // let frontier = await this.querier.sct.sctFrontier(new SctFrontierRequest({ withProof: true }))
+
       // create first epoch
       await this.indexedDb.addEpoch(0n);
 
--- a/packages/query/src/queriers/sct.ts
+++ b/packages/query/src/queriers/sct.ts
@@ -5,6 +5,8 @@ import { SctQuerierInterface } from '@penumbra-zone/types/querier';
 import {
   TimestampByHeightRequest,
   TimestampByHeightResponse,
+  SctFrontierRequest,
+  SctFrontierResponse,
 } from '@penumbra-zone/protobuf/penumbra/core/component/sct/v1/sct_pb';
 
 export class SctQuerier implements SctQuerierInterface {
@@ -13,7 +15,12 @@ export class SctQuerier implements SctQuerierInterface {
   constructor({ grpcEndpoint }: { grpcEndpoint: string }) {
     this.client = createClient(grpcEndpoint, SctService);
   }
+
   timestampByHeight(req: TimestampByHeightRequest): Promise<TimestampByHeightResponse> {
     return this.client.timestampByHeight(req);
+  }
+
+  sctFrontier(req: SctFrontierRequest): Promise<SctFrontierResponse> {
+    return this.client.sctFrontier(req);
   }
 }


apps/extension/src/message/listen/internal-services.ts
```diff
@@ -25,6 +25,14 @@ export const internalServiceListener = (
...
           indexedDb.clear(),
           localExtStorage.remove('fullSyncHeight'),
+
+          // Side-effect of resetting the cache strips the database and resets the user state.
+          // This flag gates whether the wallet pulls the 'latest' state commitment tree frontier
+          // from the full node, and removing this flag is neccessary to prevent that. On cache
+          // resets during its lifecycle, the wallet will still use the 'walletCreationBlockHeight'
+          // flag (canonically refered to as the wallet birthday) to fall back to the original
+          // sync acceleration affordances provided by that flag.
+          localExtStorage.remove('compactFrontierBlockHeight'),
```
TalDerei, line 35: comment: without removing this flag from local extension storage on cache resets, it will try and refetch the latest frontier snapshot from the full node. This is meant to serve as a one-time initialization flag

packages/query/src/block-processor.ts
```diff
@@ -167,6 +172,50 @@ export class BlockProcessor implements BlockProcessorInterface {
       { retry: () => true },
     );
 
+    // Check that 'currentHeight' and 'compactFrontierBlockHeight' local extension
+    // storage parameters match, signifying that this wallet was freshly generated
+    // and triggering a one-time parameter initialization that would have otherwise occured
+    // from fields pulled directly from the compact blocks. Otherwise, current height
+    // is set to 'PRE_GENESIS_SYNC_HEIGHT' which will set of normal genesis syncing.
+
+    if (this.compactFrontierBlockHeight && this.compactFrontierBlockHeight >= currentHeight) {
```
TalDerei, line 181: comment: a one-time operation for fresh wallets in the block processor.

packages/query/src/block-processor.ts
```diff
@@ -167,6 +172,50 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+          new FmdParameters({
+            precisionBits: 0,
+            asOfBlockHeight: appParams.shieldedPoolParams.fmdMetaParams.fmdGracePeriodBlocks,
+          }),
+        );
+      }
+
+      // Finally, persist the frontier to IndexedDB.
+      let flush = this.viewServer.flushUpdates();
+      await this.indexedDb.saveScanResult(flush);
```
TalDerei, line 216: comment: not strictly necessary since this will flush during processing the first compact block, but we persist it as a good practice.

packages/context/src/index.ts
```diff
@@ -39,50 +40,14 @@ export class Services implements ServicesInterface {
...
 
-  /**
-   * Attempt to fetch parameters from the remote fullnode, or fall back to known
-   * parameters in indexedDb.
-   *
-   * Will throw to abort if the remote node reports an unexpected chainId.
-   *
-   * @returns `AppParameters`
-   */
-  private async getParams(indexedDb: IndexedDb, querier: RootQuerier): Promise<AppParameters> {
```
TalDerei, line 50: comment: this wasn't used anywhere afaict, cc @turbocrime

TalDerei:
note to reviewers: this is now ready for review following some organizational changes.

conorsch:
> @conorsch before this lands, we should check why attempting to call the SCT frontier service on mainnet returns ConnectError: [unimplemented], but succeeds on testnet. Otherwise, syncing will break.

SCT frontier was added in [v1.5.0](https://github.com/penumbra-zone/penumbra/releases/tag/v1.5.0), roughly two weeks ago. If the RPC you're using in Prax has not upgraded to v1.5.0 or later, then the SCT frontier calls won't be available. We cannot assume that nodes will upgrade on any specific schedule, with the possible exception of coordinated chain upgrades, so the functionality in Prax should gracefully fail. It's not acceptable that "syncing will break" if the SCT frontier call is not handled.

TalDerei:
> SCT frontier was added in [v1.5.0](https://github.com/penumbra-zone/penumbra/releases/tag/v1.5.0), roughly two weeks ago. If the RPC you're using in Prax has not upgraded to v1.5.0 or later, then the SCT frontier calls won't be available. We cannot assume that nodes will upgrade on any specific schedule, with the possible exception of coordinated chain upgrades, so the functionality in Prax should gracefully fail. It's not acceptable that "syncing will break" if the SCT frontier call is not handled.

this PR will be blocked until these pd version upgrades are socialized

apps/extension/src/hooks/full-sync-height.ts
```diff
@@ -19,10 +17,10 @@ const tryGetMax = (a?: number, b?: number): number | undefined => {
 // There is a slight delay with Zustand loading up the last block synced.
 // To prevent the screen flicker, we use a loader to read it from chrome.storage.local.
 const useFullSyncHeight = (): number | undefined => {
+  // Always fetch the block height from storage to indicate current sync progress.
   const { fullSyncHeight: localHeight } = useLoaderData() as PopupLoaderData;
```
TalDerei, line 21: comment: since we're now saving the `fullSyncHeight` immediately after onboarding instead of within the block processor, there was a subtle bug where prax would sometimes momentarily display a high sync height after onboarding before self-correcting. This didn’t impact functionality.

---------

image_720

erwanor:
Can we add a hint to the error message that you should change RPC?

TalDerei:
> Can we add a hint to the error message that you should change RPC?

successful onboarding competition constitutes a level of confidence in the reliability of the selected rpc, so that shouldn't be necessary. It currently auto samples rpcs from the registry until it finds one that works. https://github.com/prax-wallet/prax/pull/336/commits/dcc94c5c1abf3eb6e595d9edf775d2e5a6610ba5 also adds fallback initialization support in case fetching the frontier from the full node fails for any reason.

apps/extension/src/hooks/full-sync-height.ts
TalDerei: womp; it doesn't seem to fix it afaict, zustand rendering is strange

TalDerei:
happy with the implementation, good to merge once it gets the green light 👍

conorsch:
RC helpfully pinged the community RPC maintainers, and most of them are now serving `v1.5.1`, which is new enough to have Frontier support. Still lagging is Crouton but RC says that'll be handled momentarily; Simply Staking's endpoint is down, so I'd prefer removing them from the registry settings temporarily until that's resolved.

conorsch (approved):

PR #339, opened 2025-05-08
block processor: process incentivized asset 

TalDerei:
the block processor improperly prevented votes from being saved to storage if the metadata wasn't available, and it no longer makes that assumption. It instead saves the incentivized asset id to storage directly during vote processing, rather than relying on the full metadata.

@turbocrime this partial indexedDB schema modification _shouldn't_ require bumping the idb version since it doesn't break backwards compat:
- the field is intentionally omitted in future writes,
- there are no existing consumers relying on the older records that include it

requires bumping storage npm deps to pass CI

pairs with https://github.com/penumbra-zone/web/pull/2337


packages/query/src/block-processor.ts
```diff
@@ -649,11 +644,11 @@ export class BlockProcessor implements BlockProcessorInterface {
 
       // One DB save per delegation asset ID, potentially spanning multiple actions within the same transaction.
       // Initially, the voting reward will be empty.
-      if (incentivizedAssetMetadata) {
+      if (incentivizedAsset) {
```
TalDerei, line 647: comment: only save vote to storage if asset Id (rather than entire metadata) exists since it can be undefined.

packages/query/src/block-processor.ts
```diff
@@ -621,13 +622,10 @@ export class BlockProcessor implements BlockProcessorInterface {
...
-          const incentivizedTokenAssetId = new AssetId({
-            altBaseDenom: action.value.body.incentivized?.denom,
-          });
-
-          incentivizedAssetMetadata =
-            await this.indexedDb.getAssetsMetadata(incentivizedTokenAssetId);
+        const denom = action.value.body.incentivized?.denom;
+        if (denom) {
+          const hash = await sha256Hash(new TextEncoder().encode(denom));
+          incentivizedAsset = new AssetId({ inner: hash });
```
TalDerei, line 628: Screenshot 2025-05-08 at 3 43 28 PM

cronokirby (approved):
Optimistically, lints should go away when we depend on the current main of web

packages/query/src/block-processor.ts
```diff
@@ -707,9 +702,9 @@ export class BlockProcessor implements BlockProcessorInterface {
             // As a result, the corresponding rewards will be denominated separately, using the
             // respective asset ID associated with each delegation's validator.
             await this.indexedDb.saveLQTHistoricalVote(
+              existingVote.incentivizedAsset,
```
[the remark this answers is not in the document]
cronokirby: Actually the write order, was changed upstream

PR #341, opened 2025-05-10
onboarding: redirect to veil

TalDerei:
references https://github.com/penumbra-zone/web/issues/2173

streamlines onboarding flow to veil by eliminating intermediate loading screens. the motivation is now that portfolio v2 supports direct IBC deposits / withdrawals into the MASP, users are able to trade directly in veil without relying on minifront.
 cc @garwalsh @VanishMax 

--------------

_before_

https://github.com/user-attachments/assets/03fb25b2-6f31-471a-9d97-7ecd9f11c654

-------------

_after_

https://github.com/user-attachments/assets/20b3052c-3c53-4e33-aad2-69de022e9520


apps/extension/src/routes/page/onboarding/success.tsx
```diff
@@ -1,26 +1,18 @@
...
-  );
+  useEffect(() => {
+    if (!hasOpened.current && defaultFrontendUrl) {
+      hasOpened.current = true;
+      window.open(defaultFrontendUrl, '_blank');
+      window.close();
+    }
+  }, [defaultFrontendUrl]);
+
+  return null;
```
TalDerei, line 17: returning undefined here was the only way to tell react that this is a valid component and render nothing. think it's fine?

TalDerei: our strict linter was complaining

PR #345, opened 2025-05-14
epoch processing consistency

TalDerei:
In line with the spirit of https://github.com/prax-wallet/prax/pull/324 that opts for a more minimal diff:
- excludes atomic validator status updates which will be handled separately (eg. https://github.com/penumbra-zone/web/pull/2258 will be stripped down accordingly),
- preserves the special casing for genesis sync,
- moves `handleEpochTransition` directly into the main block-processing loop

depends on https://github.com/penumbra-zone/web/pull/2242

### DIFF
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -55,6 +55,14 @@ declare global {
   var __ASSERT_ROOT__: boolean | undefined;
 }
 
+const PRE_GENESIS_SYNC_HEIGHT = -1n;
+
+// Chunking policy is ~3x larger than the maximum number of state payloads that are supported
+// inside a compact block (~100KB). We don't neccessarily need to determine the optimal chunk
+// size, rather it needs to be sufficiently small that devices can handle crossing the syncing
+// hurdle of processing the genesis block.
+const GENESIS_CHUNK_SIZE = 500;
+
 const isSwapRecordWithSwapCommitment = (
   r?: unknown,
 ): r is Exclude<SwapRecord, { swapCommitment: undefined }> =>
@@ -121,6 +129,16 @@ export class BlockProcessor implements BlockProcessorInterface {
     this.compactFrontierBlockHeight = compactFrontierBlockHeight;
   }
 
+  private async *streamBlocks(currentHeight: bigint) {
+    let startHeight = currentHeight + 1n;
+
+    yield* this.querier.compactBlock.compactBlockRange({
+      startHeight,
+      keepAlive: true,
+      abortSignal: this.abortController.signal,
+    });
+  }
+
   // If sync() is called multiple times concurrently, they'll all wait for
   // the same promise rather than each starting their own sync process.
   public sync = (): Promise<void> =>
@@ -156,17 +174,11 @@ export class BlockProcessor implements BlockProcessorInterface {
    * - iterate
    */
   private async syncAndStore() {
-    const PRE_GENESIS_SYNC_HEIGHT = -1n;
-
-    // Chunking policy is ~3x larger than the maximum number of state payloads that are supported
-    // inside a compact block (~100KB). We don't neccessarily need to determine the optimal chunk
-    // size, rather it needs to be sufficiently small that devices can handle crossing the syncing
-    // hurdle of processing the genesis block.
-    const GENESIS_CHUNK_SIZE = 500;
-
     // start at next block, or genesis if height is undefined
     const fullSyncHeight = await this.indexedDb.getFullSyncHeight();
     let currentHeight = fullSyncHeight ?? PRE_GENESIS_SYNC_HEIGHT;
+    let currentEpoch =
+      currentHeight && (await this.indexedDb.getEpochByHeight(currentHeight)).index;
 
     // this is the first network query of the block processor. use backoff to
     // delay until network is available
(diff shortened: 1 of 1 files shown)


packages/query/src/block-processor.ts
```diff
@@ -331,11 +368,31 @@ export class BlockProcessor implements BlockProcessorInterface {
         skipTrialDecrypt,
       });
 
+      // The presence of `epochRoot` indicates that this is the final block of the current epoch.
+      if (compactBlock.epochRoot) {
+        currentEpoch++;
+        // currentHeight++; // todo: think about this.
```
TalDerei, line 374: comment: remove, we already set `currentHeight = compactBlock.height`

packages/query/src/block-processor.ts
```diff
@@ -331,11 +368,31 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+      }
+
       // We only query Tendermint for the latest known block height once, when
       // the block processor starts running. Once we're caught up, though, the
       // chain will of course continue adding blocks, and we'll keep processing
       // them. So, we need to update `latestKnownBlockHeight` once we've passed
       // it.
+      //
+      // todo: think about moving this to top of loop and downstream side-effects
+      // like flush checks being false.
```
TalDerei, line 395: comment: this is something I've been pondering. It would be useful to think about where `FULL_SYNC_HEIGHT` that's being persisted to storage is currently being used.

packages/query/src/block-processor.ts
```diff
@@ -909,10 +957,14 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+    nextEpochStart: bigint,
+    nextHeightStart: bigint,
   ): Promise<void> {
-    const nextEpochStartHeight = endHeightOfPreviousEpoch + 1n;
-    await this.indexedDb.addEpoch(nextEpochStartHeight);
+    await this.indexedDb.addEpoch({ index: nextEpochStart, startHeight: nextHeightStart });
 
+    // TODO: seperately remove the estimation happening here and make
+    // validator status updates actually atomic using indexedDB method.
+    const nextEpochStartHeight = endHeightOfPreviousEpoch + 1n;
```
TalDerei, line 967: comment: should happen after https://github.com/penumbra-zone/web/pull/2360 lands. cut out to reduce scope creep

packages/query/src/block-processor.ts
```diff
@@ -121,6 +132,16 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+  private async *streamBlocks(currentHeight: bigint) {
+    let startHeight = currentHeight + 1n;
+
+    yield* this.querier.compactBlock.compactBlockRange({
+      startHeight,
+      keepAlive: true,
+      abortSignal: this.abortController.signal,
+    });
+  }
+
```
TalDerei, line 144: comment: same `streamBlocks` helper without the genesis check since we're still special casing

packages/query/src/block-processor.ts
```diff
@@ -156,18 +177,24 @@ export class BlockProcessor implements BlockProcessorInterface {
...
     // start at next block, or genesis if height is undefined
     const fullSyncHeight = await this.indexedDb.getFullSyncHeight();
     let currentHeight = fullSyncHeight ?? PRE_GENESIS_SYNC_HEIGHT;
 
+    // detects if this is a freshly created wallet that skipped historical sync
+    // by comparing the sync progress with the compact block frontier height.
+    const isFreshWallet =
+      this.compactFrontierBlockHeight !== undefined &&
+      fullSyncHeight !== undefined &&
+      this.compactFrontierBlockHeight >= currentHeight;
```
TalDerei, line 189: comment: https://github.com/prax-wallet/prax/pull/336/files/b4774cf6c6653d71eff0380916aec82211cb72ab#diff-ccb0493c986b958cf859e773170457e5ff59a4a2e0d742698492247f6ae872daR175-R179

packages/query/src/block-processor.ts
```diff
@@ -156,18 +177,24 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+
+    // if we're not pre-genesis and not a fresh wallet, fetch the epoch from IndexedDB.
+    // otherwise, default to epoch 0.
+    let currentEpoch =
+      currentHeight !== PRE_GENESIS_SYNC_HEIGHT && !isFreshWallet
+        ? (await this.indexedDb.getEpochByHeight(currentHeight))!.index
```
TalDerei, line 195: comment: this doesn't seem like the cleanest approach? for fresh wallets or during genesis sync, we temporarily set `currentEpoch` to `0n`, and later override it [below](https://github.com/prax-wallet/prax/pull/345/files#diff-ccb0493c986b958cf859e773170457e5ff59a4a2e0d742698492247f6ae872daR216-R222); otherwise, we fetch it from storage.

packages/query/src/block-processor.ts
```diff
@@ -186,12 +213,22 @@ export class BlockProcessor implements BlockProcessorInterface {
     // and triggering a one-time parameter initialization that would have otherwise occured
     // from fields pulled directly from the compact blocks. Otherwise, current height
     // is set to 'PRE_GENESIS_SYNC_HEIGHT' which will set of normal genesis syncing.
+    if (isFreshWallet) {
+      // One of the neccessary steps here to seed the database with the first epoch
+      // is a one-time request to fetch the epoch for the corresponding frontier
+      // block height from the full node, and save it to storage.
+      const remoteEpoch = await this.querier.sct.epochByBlockHeight(
+        new EpochByHeightRequest({ height: currentHeight }),
+      );
```
TalDerei, line 222: comment: one-time call on wallet init to fetch epoch from full node and persist to db

JasonMHasperhoven:
There are some lint errors

packages/query/src/block-processor.ts
turbocrime: discussion in https://github.com/prax-wallet/prax/pull/324#discussion_r2047180261 covers my thoughts on how the loop should manage this

turbocrime: imo, FULL_SYNC_HEIGHT in the database should either be removed or accessed only once. it's maintained in extension storage also

PR #346, opened 2025-05-15
block processor: fetch remote epoch

TalDerei:
persists remote epoch to storage and seeds validator info

pairs with https://github.com/penumbra-zone/web/pull/2372 and requires consuming latest npm deps to pass CI


packages/query/src/block-processor.ts
```diff
@@ -168,6 +171,13 @@ export class BlockProcessor implements BlockProcessorInterface {
     const fullSyncHeight = await this.indexedDb.getFullSyncHeight();
     let currentHeight = fullSyncHeight ?? PRE_GENESIS_SYNC_HEIGHT;
 
+    // detects if this is a freshly created wallet that skipped historical sync
+    // by comparing the sync progress with the compact block frontier height.
+    const isFreshWallet =
+      this.compactFrontierBlockHeight &&
+      fullSyncHeight !== undefined &&
+      this.compactFrontierBlockHeight >= currentHeight;
```
TalDerei, line 179: comment: pulled conditional into a var

erwanor (approved):

turbocrime (approved):

packages/query/src/block-processor.ts
```diff
@@ -909,9 +934,11 @@ export class BlockProcessor implements BlockProcessorInterface {
   private async handleEpochTransition(
     endHeightOfPreviousEpoch: bigint,
     latestKnownBlockHeight: bigint,
+    epochIndex: bigint,
   ): Promise<void> {
     const nextEpochStartHeight = endHeightOfPreviousEpoch + 1n;
-    await this.indexedDb.addEpoch(nextEpochStartHeight);
+
+    await this.indexedDb.addEpoch(nextEpochStartHeight, epochIndex);
```
turbocrime, line 941: epochIndex is possibly off-by-one here

turbocrime:
```suggestion
    const nextEpochIndex = epochIndex + 1n;
    await this.indexedDb.addEpoch(nextEpochStartHeight, nextEpochIndex);
```

PR #353, opened 2025-05-22
fix: content scripts attempt to connect during prerender

turbocrime:
speculative navigation may cause content scripts to execute before a document is active. at this point, the document `sender` can't be validated by the requirements of the extension service worker.

this PR uses the document's `prerendering` attribute to detect this state, and the `prerenderingchange` event to detect transition out of this state.

### DIFF
--- a/apps/extension/src/content-scripts/message/prerendering.ts
+++ b/apps/extension/src/content-scripts/message/prerendering.ts
@@ -1,0 +1,6 @@
+/** While prerendering, the document isn't a valid sender. */
+export const prerendering = new Promise<void>(resolve =>
+  document.prerendering
+    ? document.addEventListener('prerenderingchange', () => resolve(), { once: true })
+    : resolve(),
+);
--- a/apps/extension/src/content-scripts/message/send-background.ts
+++ b/apps/extension/src/content-scripts/message/send-background.ts
@@ -1,9 +1,41 @@
 import { PenumbraRequestFailure } from '@penumbra-zone/client';
-import type { PraxConnection } from './prax-connection';
+import { PraxConnection } from './prax-connection';
+import { prerendering } from './prerendering';
 
-export const sendBackground = async (
+/**
+ * Private lock, only shared locally among importers in the same runtime.
+ * @note Can't use `navigator.locks` because it would escape the content script.
+ */
+let lock: Promise<null | PenumbraRequestFailure> = Promise.resolve(null);
+let steal = new AbortController();
+
+/**
+ * Send a connection lifecycle message to the extension service worker.
+ *
+ * @note not async to avoid reentrancy
+ */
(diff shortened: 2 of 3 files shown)


apps/extension/src/content-scripts/message/send-background.ts
```diff
@@ -1,9 +1,41 @@
...
+import { PraxConnection } from './prax-connection';
+import { prerendering } from './prerendering';
 
-export const sendBackground = async (
+/**
+ * Private lock, only shared locally among importers in the same runtime.
+ * @note Can't use `navigator.locks` because it would escape the content script.
+ */
+let lock: Promise<null | PenumbraRequestFailure> = Promise.resolve(null);
+let steal = new AbortController();
```
turbocrime, line 10: significant functional change. a lock queue is added to preserve message order.

this is implemented here because navigator locks would be available globally to any script executing on the origin, which isn't acceptable in a content script that wants to remain isolated.

apps/extension/src/content-scripts/message/send-background.ts
```diff
@@ -1,9 +1,41 @@
...
+    lock.catch().then(() => {
+      stolen.throwIfAborted();
+      return innerSendBackground(message);
+    }));
+};
+
+const innerSendBackground = async (
   message: PraxConnection,
 ): Promise<null | PenumbraRequestFailure> => {
+  await prerendering;
```
turbocrime, line 38: waits for prerendering to finish before emitting any message.

apps/extension/src/content-scripts/message/prerendering.ts
```diff
@@ -0,0 +1,6 @@
+/** While prerendering, the document isn't a valid sender. */
+export const prerendering = new Promise<void>(resolve =>
+  document.prerendering
+    ? document.addEventListener('prerenderingchange', () => resolve(), { once: true })
+    : resolve(),
+);
```
turbocrime, line 6: this is a constant, created on script load and resolving once.

if the document never renders, it will never resolve.

apps/extension/src/content-scripts/message/send-background.ts
```diff
@@ -1,9 +1,41 @@
...
+
+  // capture present signal within the closure
+  const stolen = steal.signal;
+
+  return (lock = // update the lock
+    // wait for turn without failure
+    lock.catch().then(() => {
+      stolen.throwIfAborted();
+      return innerSendBackground(message);
+    }));
```
turbocrime, line 32: lock logic. this implementation was chosen because it's simple and small.

TalDerei: what if `chrome.runtime.sendMessage()` hangs, freezing the queue?

turbocrime: not likely, and anyway, there's not much we can do about that. disconnect is all one might really want to do, which can take over in that case.

PR #356, opened 2025-05-26
feat: allow content scripts to init during prerender

turbocrime:
improvement/alternative solution to #353


apps/extension/src/content-scripts/message/send-background.ts
```diff
@@ -1,28 +1,70 @@
...
+/**
+ * Private lock, only shared locally among importers in the same runtime (the
+ * document hosting this isolated content script).
+ *
+ * The initial lock 'holder' is the prerendering state, which must complete
+ * before communication is permitted.
+ *
+ * @note Can't use `navigator.locks` because it would escape the content script.
+ */
+let bgLock: Promise<null | PenumbraRequestFailure> = new Promise<null>(resolve =>
```
TalDerei, line 14: so in theory, the sequential execution queue should make extension context invalidated errors (and more broadly messages that are sent but never received) less frequent and more predictable by reducing possible race conditions?

apps/extension/src/content-scripts/message/send-background.ts
```diff
@@ -1,28 +1,70 @@
...
+  // disconnect can jump the queue
+  if (message === PraxConnection.Disconnect) {
+    bgSteal.abort();
+    bgSteal = new AbortController();
+  }
+
+  // capture present signal within the closure
+  const stolen = bgSteal.signal;
+
+  return (bgLock = // update the lock
```
TalDerei, line 39: what if chrome.runtime.sendMessage() hangs, freezing the queue? is praxConnection.Disconnect a sufficient escape hatch here?

apps/extension/src/content-scripts/message/send-background.ts
turbocrime: maybe. it does prevent rapid requests from creating some kinds of race conditions, but there is no reason or example of a caller making rapid requests, and there weren't any concerning cases.

apps/extension/src/content-scripts/message/send-background.ts
TalDerei: is it necessary to clear the queue after lock stealing?

TalDerei: let's push it further, what if the service worker dies while processing disconnect?

TalDerei: we have session-driven disconnects, but if the queue is frozen, then how can it process `sendBackground(PraxConnection.Disconnect)`, given that it also uses `chrome.runtime.sendMessage()`?

TalDerei: eg. If chrome arbitrarily kills the service worker, the content script will wait indefinitely because the promise returned by `chrome.runtime.sendMessage()` can hang, even across service worker reloads, right? Holistically, we seem to have traded race condition issues for a hanging queue problem.

PR #368, opened 2025-06-02
reset password: re-encrypt seed phrase after setting new password

TalDerei:
https://github.com/prax-wallet/prax/pull/361 originally implemented the "reset password" feature, but the implementation was incomplete:

- a new key was derived from the new password and stored as `KeyJson` in memory and session storage,
- the corresponding `KeyPrint` was saved to local storage for future password validation.

However, it _did not_ re-encrypt the existing seed phrase with the new password-derived key after resetting the password. This implementation now does that by (1) decrypting the existing seed using the existing key, (2) setting the new password, and (3) re-encrypting the seed phrase using the new password key.

### DIFF
--- a/apps/extension/src/routes/popup/settings/settings-reset-password.tsx
+++ b/apps/extension/src/routes/popup/settings/settings-reset-password.tsx
@@ -5,15 +5,17 @@ import { useStore } from '../../../state';
 import { passwordSelector } from '../../../state/password';
 import { SettingsScreen } from './settings-screen';
 import { KeyGradientIcon } from '../../../icons/key-gradient';
+import { walletsSelector } from '../../../state/wallets';
 
 export const SettingsResetPassword = () => {
-  const { isPassword, setPassword: saveNewPassword } = useStore(passwordSelector);
-
   const [currentPassword, setCurrentPassword] = useState('');
   const [newPassword, setNewPassword] = useState('');
   const [confirmPassword, setConfirmPassword] = useState('');
   const [enteredIncorrect, setEnteredIncorrect] = useState(false);
   const [saved, setSaved] = useState(false);
+  const { isPassword, setPassword } = useStore(passwordSelector);
+  const { reencryptSeedPhrase } = useStore(walletsSelector);
+  const { getSeedPhrase } = useStore(walletsSelector);
 
   const submit = (e: React.FormEvent<HTMLFormElement>) => {
     e.preventDefault();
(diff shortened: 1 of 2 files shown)


apps/extension/src/routes/popup/settings/settings-reset-password.tsx
```diff
@@ -25,11 +27,14 @@ export const SettingsResetPassword = () => {
...
 
-      if (newPassword !== confirmPassword || newPassword.length < 8) {
+      if (newPassword !== confirmPassword) {
         return;
       }
 
-      await saveNewPassword(newPassword);
+      const seedPhrase = await getSeedPhrase();
+      await setPassword(newPassword);
+      await reencryptSeedPhrase(seedPhrase);
```
TalDerei, line 36: `passwordKeyPrint` and `encryptedSeedPhrase` in LS should change after resetting password

--------------

Screenshot 2025-06-02 at 12 54 39 PM

Screenshot 2025-06-02 at 12 53 50 PM

apps/extension/src/state/wallets.ts
```diff
@@ -63,6 +64,35 @@ export const createWalletsSlice =
...
+        const key = await Key.fromJson(passwordKey);
+        const encryptedSeedPhrase = await key.seal(seedPhraseStr);
+
+        // Update in-memory store
+        set(state => {
+          const wallet = state.wallets.all[0];
+          if (!wallet) {
+            throw new Error('No wallet found in state');
+          }
+          wallet.custody.encryptedSeedPhrase = encryptedSeedPhrase;
```
cronokirby, line 84: I don't think this works with how zustand operates, we need to instead return a new state object with this modified, right?

apps/extension/src/state/wallets.ts
```diff
@@ -63,6 +64,35 @@ export const createWalletsSlice =
...
+        });
+
+        // Update local storage
+        const wallets = await local.get('wallets');
+        if (!wallets.length || !wallets[0]) {
+          throw new Error('No wallets in local storage');
+        }
+
+        wallets[0].custody.encryptedSeedPhrase = encryptedSeedPhrase.toJson();
```
cronokirby, line 93: same here

cronokirby (changes requested):
I don't think the in-memory store update is correct as written, and we should also handle updating all the wallets, rather than just hard-coding in wallet 0. (We can figure out if there's a relevant encrypted seed phrase dynamically, and the first wallet might have been made with ledger).

Also, we should avoid duplicating non-trivial logic like this, let's at least extract that out into a common function.

apps/extension/src/state/wallets.ts
```diff
@@ -63,6 +64,35 @@ export const createWalletsSlice =
...
+          if (!wallet) {
+            throw new Error('No wallet found in state');
+          }
+          wallet.custody.encryptedSeedPhrase = encryptedSeedPhrase;
+        });
+
+        // Update local storage
+        const wallets = await local.get('wallets');
+        if (!wallets.length || !wallets[0]) {
+          throw new Error('No wallets in local storage');
```
cronokirby, line 90: This can just be a no-op

TalDerei:
closing as unplanned

PR #369, opened 2025-06-02
temporarily disable reset password feature

TalDerei:
precursor to eventual flushed out implementation https://github.com/prax-wallet/prax/pull/368


conorsch (approved):
Temporary reversion, intended to unblock shipping. :saluting_face:

PR #370, opened 2025-06-11
use navigator lock for storage access

turbocrime:
instead of using a document-local promise, use an origin-global navigator lock when performing migrations and storage access.

### DIFF
--- a/apps/extension/package.json
+++ b/apps/extension/package.json
@@ -1,6 +1,6 @@
 {
   "name": "chrome-extension",
-  "version": "19.1.0",
+  "version": "19.2.0",
   "private": true,
   "license": "(MIT OR Apache-2.0)",
   "description": "chrome-extension",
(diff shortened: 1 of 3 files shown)


packages/storage-chrome/src/base.ts
```diff
@@ -150,24 +145,14 @@ export class ExtensionStorage<T extends { dbVersion: number }> {
...
-    }
-
-    try {
-      this.dbLock = this.migrateOrInitializeIfNeeded();
-      await this.dbLock;
-      return await fn();
-    } finally {
-      this.dbLock = undefined;
-    }
+    return navigator.locks.request(`${chrome.runtime.id}.storage`, { mode: 'exclusive' }, () =>
```
TalDerei, line 153: ACK – global exclusive lock, and all extension contexts share the same origin. replaces our bespoke locking mechanism

optionally, mark it as async for _stylistic_ purposes.

TalDerei (approved):
👍

Issue #371, opened 2025-06-11
storage: mock used for tests exhibits inaccurate function


turbocrime:
the mock used for storage tests contains incorrect behavior.

in the `set` method:

https://github.com/prax-wallet/prax/blob/c23eb76bc3eed34025b7ebe48461d3c234115d7e/packages/storage-chrome/src/mock.ts#L54-L65

setting undefined values in chrome storage is a no-op.

this is not documented, but it's demonstrable with simple experimentation.

<img width="477" alt="Image" src="https://github.com/user-attachments/assets/ece26e73-1c43-4607-952b-41669deedada" />

correcting this mock causes migration tests to fail, so migrations seem to rely upon the wrong behavior.

PR #372, opened 2025-06-11
Correct storage mock `set` behavior

turbocrime:
fixes https://github.com/prax-wallet/prax/issues/371

to correct behavior without modifying migrations, the migration utility now calls `clear` before calling `set` to apply the migrated data.

### DIFF
--- a/packages/storage-chrome/src/mock.ts
+++ b/packages/storage-chrome/src/mock.ts
@@ -39,6 +39,8 @@ export class MockStorageArea implements chrome.storage.StorageArea {
       const value = this.store.get(key);
       if (value !== undefined) {
         result[key] = value;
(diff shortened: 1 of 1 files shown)


packages/storage-chrome/src/base.ts
```diff
@@ -183,6 +170,7 @@ export class ExtensionStorage<T extends { dbVersion: number }> {
     // Migrations save the database intermediate states hard to type
     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- EXISTING USE
     const nextState = await migrationFn(currentDbState);
+    await this.storage.clear();
```
turbocrime, line 173: due to concern about storage conflicts with cross-versioned documents, i think use of `clear` during migration is dangerous and should be avoided.

PR #385, opened 2025-06-25
ci: don't create draft release objects

conorsch:
Partial reversion of 5b01bf98bf9e74eb421bb8b829148cdbdfe7c7df, from PR #378. Keeps the relocated GHA references, drops the draft release object, as we prefer the old-style changelogs.

## Screenshot of draft release object
gh-release

## Here's that same release object edited:

release-final

### DIFF
--- a/.github/workflows/extension-upload.yml
+++ b/.github/workflows/extension-upload.yml
@@ -43,22 +43,6 @@ jobs:
       - name: Zip extension versions
         run: pnpm zip
 
-      - name: Draft preview release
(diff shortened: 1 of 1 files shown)


TalDerei (commented):
Is possible to have expose both release artifacts?

TalDerei (approved):

turbocrime:
reverting is fine with me, but i'm wondering why a release asset wasn't created by #378?

turbocrime (approved):

PR #389, opened 2025-06-25
block processor: decrypt position metadata in the block processor

TalDerei:
references https://github.com/penumbra-zone/web/issues/2502

CI failing until updated npm deps are consumed.


packages/query/src/block-processor.ts
```diff
@@ -877,22 +890,44 @@ export class BlockProcessor implements BlockProcessorInterface {
...
         });
       }
+
+      // Optionally, decrypt position metadata by calling TS wrapper that invokes wasm decryption method.
+      let decryptedPositionMetadata: PositionMetadata | undefined;
+      if (action.value.encryptedMetadata.length !== 0) {
+        try {
+          decryptedPositionMetadata = decryptPositionMetadata(
+            this.fullViewingKey,
+            action.value.encryptedMetadata,
```
TalDerei, line 900: comment: passing full viewing key from `ProcessBlockParams` to decrypt the position metadata

PR #384, opened 2025-06-24
refactor (storage): improve wrapper, improve migration tools

turbocrime:
pending https://github.com/prax-wallet/prax/pull/386

- Use normal `chrome.storage.StorageArea` type instead of arbitrary `IStorage`
- Storage wrapper types improved
- Storage wrapper race conditions eliminated
- Storage wrapper applies migrations in a safer way
- Migrations and version definitions are more constrained in format to support correctness

PR #364, opened 2025-05-30
onboarding: redirect to landing page

TalDerei:
precursor to https://github.com/prax-wallet/prax/pull/365 that will be folded in.

--------------

https://github.com/user-attachments/assets/146e7dd9-8e6e-4e5f-834e-7fdcc8bd11f8

PR #390, opened 2025-07-03
refactor: collect record types in storage package, introduce internal encryption and wallet packages

turbocrime:
pending #384

- Storage record types (`OriginRecord`, `UserChoice`) reside in `@repo/storage-chrome`
- Password and encryption utilities are collected in `@repo/encryption`, removing dependency on `@penumbra-zone/crypto-web`
- Wallet is in a dedicated package at `@repo/wallet` to scope pending refactor
- Relocate mnemonic functionality to internal implementation
- Remove disabled settings components (reset password)


apps/extension/src/ctx/spend-key.test.ts
```diff
@@ -1,27 +1,10 @@
...
-  return {
-    ...mod,
-    sessionExtStorage,
-  };
-});
-
-// needs to be imported after storage mocks are set up
-const { getSpendKey } = await import('./spend-key');
+const localMock = (chrome.storage.local as unknown as { mock: Map<string, unknown> }).mock;
+const sessionMock = (chrome.storage.session as unknown as { mock: Map<string, unknown> }).mock;
```
TalDerei, line 7: can we rebase on top of current main to reduce diff. captured in https://github.com/prax-wallet/prax/pull/381

apps/extension/src/routes/popup/settings/routes.tsx
```diff
@@ -33,16 +32,8 @@ export const settingsRoutes = [
...
-    element: <SettingsPassphrase />,
-  },
   {
     path: PopupPath.SETTINGS_NUMERAIRES,
     element: <SettingsNumeraires />,
   },
-  // {
-  //   path: PopupPath.RESET_PASSWORD,
-  //   element: <SettingsResetPassword />,
-  // },
```
TalDerei, line 47: comment: ACK – treating https://github.com/prax-wallet/prax/pull/368 as unplanned for now. 

mapping out the current state of play for onboarding: https://github.com/prax-wallet/prax/pull/364 reflects the current onboarding flow for new users (notably, the seed phrase is not displayed – the only access point at which the user can discover their recovery phrase is by navigating to the settings post-onboarding completion). Eventually, we should remove the encrypted seed phrase from extension local storage entirely, which eliminates the recovery option as a side-effect. The user isn't currently shown the seed phrase during onboarding, and won't (eventually) be able to view / save it from settings if we go down that route by removing the encrypted seed phrase from LS and disabling this route. In that context, the seed phrase becomes effectively undiscoverable, so supporting the latter would be contingent on have some kind of keychain support.

apps/extension/src/routes/popup/settings/routes.tsx
```diff
@@ -33,16 +32,8 @@ export const settingsRoutes = [
     path: PopupPath.SETTINGS_CONNECTED_SITES,
     element: <SettingsConnectedSites />,
   },
-  {
-    path: PopupPath.SETTINGS_RECOVERY_PASSPHRASE,
-    element: <SettingsPassphrase />,
-  },
```
TalDerei, line 39: question: this is not a disabled settings component, why are we removing this?

TalDerei (commented):
I'd prefer to hold off on merging until after the migration lands and we cut the next release. After that, we can open a corresponding web PR to handle any cleanup once this is included in the subsequent release.

apps/extension/src/routes/popup/settings/routes.tsx
turbocrime: recording discussion:

tl;dr this PR and https://github.com/prax-wallet/prax/pull/391 consist of a larger set of work split to scope down. this fell on the wrong side of the dividing line and is premature

in that larger set of work, removal was chosen to reduce complexity and reduce refactor. i assumed intentions to move away from storing seed phrase, and required support for new custody implementations for which no seed phrase is available.

path forward is: revert removal. a separate PR landing ahead of this one will refactor passphrase settings to understand conditional availability of the seed phrase.

that work is deferred until #392 review is satisfied.

apps/extension/src/routes/popup/settings/settings-reset-password.tsx
```diff
@@ -1,93 +0,0 @@
-import { useState } from 'react';
```
TalDerei, line 1: comment: remove the uncommented route from [routes.tsx](https://github.com/prax-wallet/prax/blob/storage-relocate-records/apps/extension/src/routes/popup/settings/routes.tsx#L44-L47)

TalDerei (approved):
previously reviewed, approving now that the reset passphrase route has been reverted. We should clean up the `@penumbra-zone/crypto-web` package after the next release.

apps/extension/src/routes/popup/settings/settings-reset-password.tsx
turbocrime: removed

Issue #393, opened 2025-07-10
`syncLastBlockToStorage` may attempt to set `NaN` value to storage


turbocrime:
`syncLastBlockToStorage` may attempt to write `NaN` to storage, resulting in unexpected behavior.

notably, this attempted `NaN` write is triggered on startup if a user onboards, and then ends the session during genesis sync. prior to prax v12 c020aea764778bb410201ad2e8b7c7a0f76db735, this reliably set up a migration corruption case, creating broken extension storage if the extension upgraded to v12 before ever setting a defined value.

the utility now:

https://github.com/prax-wallet/prax/blob/ed0b2a3b989fd6a74dbd3b2ec63d17697619c461/apps/extension/src/wallet-services.ts#L62-L80

it's worth considering that this utility has remained mostly unchanged since very early in development

https://github.com/prax-wallet/prax/blob/b2a3ab71223bb0582f1e356ba2bf7f853841d8f1/packages/storage/src/chrome/syncer.ts#L4-L14

### the problem

if `indexedDb.getFullSyncHeight` returns `undefined` or the subscription provides an update containing `undefined`, the storage wrapper call becomes:

```ts
await localExtStorage.set('fullSyncHeight', NaN);
```

this does not actually write `NaN` to storage and it does not throw. it is effectively equivalent to attempting to set an `undefined` value

#### item-versioned storage wrapper

the legacy storage wrapper would create a v0-v1 migration corruption case, but work normally until migration
```ts
chrome.storage.local({ fullSyncHeight: { version: 'V2' } })
```

#### globally-versioned storage wrapper

the c020aea764778bb410201ad2e8b7c7a0f76db735 storage wrapper would perform a no-op, meaning
- the stored height could disagree with IDB height creating unexpected behavior
- any present corrupted migration value is allowed to persist

```ts
chrome.storage.local({ fullSyncHeight: undefined })
```

turbocrime:
after looking into every `set` call, i believe this is the only location at risk. unfortunately it's not possible to exclude an `NaN` value by the type system, so it has to be a runtime failure.

turbocrime:
since #384 this issue will result in a runtime error. example from https://github.com/prax-wallet/prax/pull/394#issuecomment-3066293501

<img width="1123" height="947" alt="Image" src="https://github.com/user-attachments/assets/33d5645c-53bf-455f-a94c-b9912683a24b" />

turbocrime:
[...]

PR #394, opened 2025-07-11
avoid setting NaN fullSyncHeight

turbocrime:
fixes https://github.com/prax-wallet/prax/issues/393


turbocrime:
it would be nice to have some documentation or comments on what the purpose of the sentinel is, @TalDerei

apps/extension/src/wallet-services.ts
```diff
@@ -67,13 +67,13 @@ const getChainId = async (baseUrl: string) => {
 const syncLastBlockToStorage = async ({ indexedDb }: Pick<WalletServices, 'indexedDb'>) => {
   const dbHeight = await indexedDb.getFullSyncHeight();
 
-  if (!isSentinel(dbHeight)) {
+  if (dbHeight != null && dbHeight !== SENTINEL_U64_MAX) {
```
TalDerei, line 70: comment: why is this approach preferred? this assumes that `dbHeight` is never a number or undefined. Is that a safe assumption?

TalDerei:
I encountered this today on main syncing an existing wallet

-----------

Screenshot 2025-07-12 at 5 20 37 PM

TalDerei (approved):
this makes sense to me 👍

PR #399, opened 2025-07-23
Clean up storage/zustand persist

turbocrime:
zustand persist is slightly broken, with some misnamed keys

just drop the session persist, which never worked. use the typed listener provided by `ExtensionStorage` wrapper.

drop unused password key from zustand.

this is split out from https://github.com/prax-wallet/prax/pull/390


TalDerei (approved):
LGTM 👍

PR #401, opened 2025-07-30
Increase genesis chunking policy

TalDerei:
https://github.com/prax-wallet/prax/pull/257 introduced chunked genesis sync with a chunking policy of 500 SNRs. Through empirical testing, i've found that on particularly slow wifi connections the genesis sync often fails to complete unless you keep the extension open, possibly due to the increased frequency of wasm boundary crossings between chunking passes. Increasing the chunking policy to 2K SNRs maintains the benefits of chunking for lower-powered devices while also allowing the sync to complete reliably, even on slower connections.

references https://github.com/penumbra-zone/web/issues/2640


conorsch (approved):
Approving optimistically. =)

PR #402, opened 2025-08-02
Ledger UI: Onboarding

turbocrime:
This PR implements the user interface and onboarding flow for Ledger hardware wallet support. Building on the foundation custody infrastructure, this adds the UI components and onboarding flows. **Note: This PR has been split - authorization slice changes have been moved to PR #403.**

- Complete Ledger onboarding UI with device connection flow
- Reorganized onboarding structure to support multiple custody types  
- Enhanced password management for different wallet types

## Key Changes

- **Onboarding Flow**: New custody-specific onboarding paths for seed phrase generation, import, and Ledger connection in `apps/extension/src/routes/page/onboarding/custody/`
- **Ledger Integration**: Complete USB device pairing and app communication flow in `connect-ledger.tsx:39-88`
- **Password Flow**: Unified password creation/confirmation handling for different custody types
- **Dependencies**: Added Ledger transport and Penumbra app libraries

## Onboarding

- Device pairing flow with USB permission handling
- Wallet ID display for connected device
- Reliable error handling and user feedback for USB interruptions

PR #403, opened 2025-08-18
Ledger UI: Authorization

turbocrime:
This PR contains the authorization slice changes split out from PR #402. It implements the core transaction authorization flow updates needed to support Ledger hardware wallet signing through the custody abstraction.

Note that to access usb devices, authorization is now performed directly within the popup.

This PR focuses specifically on the authorization logic changes, while the UI components and onboarding flows remain in PR #402.


packages/wallet/src/wallet.ts
```diff
@@ -76,8 +75,16 @@ export class Wallet<T extends CustodyTypeName = CustodyTypeName> {
...
+    // const lockGrantedCallback = (async () => {
     const unsealed = await this.unseal(passKey);
     return bindCustodyImpl(this, unsealed);
+    // }) satisfies LockGrantedCallback;
+
+    // return (await navigator.locks.request(
+    //  `custody-${this.label}-${bech32mWalletId(this.id)}`,
+    //   { mode: 'exclusive' },
+    //   lockGrantedCallback,
+    // )) as never;
```
TalDerei, line 87: comment: remove

apps/extension/src/ctx/authorization.ts
```diff
@@ -1,53 +1,18 @@
...
-import { Key } from '@repo/encryption/key';
-import { localExtStorage } from '@repo/storage-chrome/local';
-import { UserChoice } from '@repo/storage-chrome/records';
-import { sessionExtStorage } from '@repo/storage-chrome/session';
-import { Wallet } from '@repo/wallet';
 import { PopupType } from '../message/popup';
-import { throwIfNeedsLogin } from '../needs-login';
 import { popup } from '../popup';
 
 export const getAuthorization = async (plan: TransactionPlan): Promise<AuthorizationData> => {
```
TalDerei, line 13: comment: general organizational question here – any downstream side-effects of relocating signing from background script to directly being owned by the popup context?

apps/extension/src/ctx/authorization.ts
```diff
@@ -1,53 +1,18 @@
...
-
-  const choose = popup(PopupType.TxApproval, {
+  const authData = await popup(PopupType.TxApproval, {
     authorizeRequest: new AuthorizeRequest({ plan }).toJson() as Jsonified<AuthorizeRequest>,
-  })
-    .then(response => response?.choice === UserChoice.Approved)
-    .catch(error => {
-      console.error(error);
-      throw new ConnectError('Approval failed', Code.Internal);
-    });
```
TalDerei, line 32: comment: if the user denies or the popup throws, popup should reject. why are we removing this, doesn't returning an error instead of rejecting crash on `response!.authorizeResponse`?

TalDerei (approved):
optimistically 👍

quite a few organizational changes here, with helpers split into separate files — but that looks fine.

